Issue #39, opened 2012-06-20
Handling of elements in Symmetric matrices


alexkaratarakis:
When working with symmetric matrices there is a potential risk of misuse because element ij and element ji is the same value. Therefore if a user makes a modification on both triangles of the matrix, the result would not be what the user expects. 
For example, user A have a 3x3 SymmetricDenseMatrix:
[0, 0, 0]
[~, 0, 0]
[~,~, 0]

Since the matrix is symmetric (s)he proceeds to add 5 to (0,2), so:
[0, 0, 5]
[~, 0, 0]
[~,~, 0]

The symmetric matrix is then expected to return 5 in both matrix[2,0] and matrix[0,2], which stands true. 

On the other hand, there is a risk of adding the value twice with (for example) the intention to preserve symmetry, If add 5 to both (0,2) and (2,0) the result is:
[0, 0, 10]
[~, 0, 0]
[~, ~, 0]

which is not what is expected. The actual cases would involve writing on both sides of the matrix for whatever reason and won't be so clear as the provided example so we have the danger of allowing silent errors which would be hard to debug.

I would like to know your opinion about how you would want to handle this in mathnet. 
My opinion, and the handling I currently use, is to disallow writing on the lower triangle of the matrix (throw an exception when attempted). Only writing on the diagonal and upper triangle is allowed, so you are alerted in case of mis-use. This has caught a huge amount of bugs.

PR #59, opened 2012-11-26
Symmetric matrices

alexkaratarakis:
Adds symmetric matrices. 

-Symmetric class family
SquareMatrix->SymmetricMatrix->SymmetricDenseMatrix

There are some choices that need to be made here:
a) The transpose of a symmetric matrix is the same as the original matrix. This means that Tranpose() can simply return "this". However, this has the pitfall of not being a separate entity and thus changes to the "transposed" affect the original. Since there was a Unit test explicitly checking for NotSameObject, Transpose() return this.clone(). 

b) Writing the symmetric matrix: [i,j] & [j,i] point to the same element. Modifying the strictly lower triangle is disabled to avoid hard to debug errors. See issue 39 (https://github.com/mathnet/mathnet-numerics/issues/39) for more information. 
- Indexer class family
  This uses the recently introduced MatrixStorage. The Storages used for Symmetric Matrices also have an Indexer as a separate class. The same indexer is used for Triangular matrices so its worth having it separately to avoid duplication. Included is a hierarchy to cover Indexers, StaticIndexers, (in contrast to DynamicIndexers used by formats such as Skyline) and PackedStorage with an implemented UpperPackedStorage.
- Commits 
  99b710a [Matrix.Arithmetic] Matrix-Matrix operations that CreateMatrix() request a fullyMutable matrix
  94747be [MatrixTests.Arithmetic] Fix reversed CanKroneckerProduct() test method names

are the only ones (I think) that modify existing files. They are bug fixes that can be independently cherry picked for master.

### DIFF
--- a/src/Numerics/LinearAlgebra/Complex32/SymmetricDenseMatrix.cs
+++ b/src/Numerics/LinearAlgebra/Complex32/SymmetricDenseMatrix.cs
@@ -233,7 +233,7 @@ namespace MathNet.Numerics.LinearAlgebra.Complex32
             var m = new SymmetricDenseMatrix(order);
             for (var i = 0; i < order; i++)
             {
-                m.At(i, i, 1.0f);
+                m.At(i, i, Complex32.One);
             }
 
             return m;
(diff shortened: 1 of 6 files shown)


Fread75:
This feature could be really useful for my use case. Do you still plan to merge it to master ?

PR #173, opened 2013-11-17
Optimization

joemoorhouse:
Optimization branch
Hi,
First go at adding in Optimization routines (intended to get the discussion started, not to be final article). Added in a wrapper around MKL non-linear least squares solver (plenty of questions there about new provider versus existing linear maths providers etc). Also added Powell method (and Brent line search) in C# to have a basic derivative-less minimizer; this is a function minimizer but can also be used to fit curves in least-squares sense. Curve fitting common for all function minimizers, so split this out. Probably need separate interfaces for function-minimizers and non-linear least squares minimizers (leaving the latter for now as possible there is only ever one algorithm).
Needs a bunch more tests...
Thanks,
Joe

### DIFF
--- a/src/NativeWrappers/MKL/optimization.cpp
+++ b/src/NativeWrappers/MKL/optimization.cpp
@@ -1,0 +1,227 @@
+#include <algorithm>
+#include <complex>
+#define MKL_Complex8 std::complex<float>
+#define MKL_Complex16 std::complex<double>
+
+#include "mkl_lapack.h"
+#include "mkl_cblas.h"
+#include "lapack_common.h"
+#include "wrapper_common.h"
+#include "mkl_lapacke.h"
+#include "mkl.h"
+#include "mkl_trans.h"
+
+template<typename T>
+inline MKL_INT lu_factor_2(MKL_INT m, T a[], MKL_INT ipiv[],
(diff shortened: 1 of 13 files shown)


forki:
Is there any progress on this?
Or is there another minimization function in Math.Net?

scottstephens:
No progress on integrating the branches. My branch continues to get bug fixes, performance improvements, etc whenever I find problems. The BFGS-B algorithm has also been added since my first comment on this thread (though it's pretty old news at this point, it was added about a year ago). BFGS-B can minimize nonlinear functions subject to upper and lower bounds on each parameter, and is suitable for small to medium size problems (I would guess up to 100 variables, but I have only used/tested it extensively on quite small problems, 3-7 variables).

eriove:
I haven't looked at the code but it would be great if it could be integrated. To me minimization is the only feature I've been missing from MathNet.

cdrnet:
Great! I'll check whether I can revive my fork of this branch which has already been migrated to v3.

scottstephens:
I've created a test branch for us to experiment on. The pull request creating it is https://github.com/mathnet/mathnet-numerics/issues/283. I've also created an issue for merging the optimization efforts, as it will require more than one pull request. That's here: https://github.com/mathnet/mathnet-numerics/issues/283

cdrnet:
This work has been merged in via another branch in the meantime. Thanks again!

Issue #196, opened 2014-02-19
Gamma function has room for improvement


verybadcat:
The gamma function produces wrong values for integer inputs.  For example, gamma(4) returns not 6, but 5.99999999999999, and gamma(5) returns not 24, but 24.00000000000011.  Similar errors happen for other integers.  I realize that some error is inevitable as doubles actually have discrete values.  But there are still two problems:

1)  For integer inputs, it really should return the correct integer output.  

2)  The error in these particular cases are respectively 14x and 32x the granularity of double near their output values.  That seems like more than it would need to be. 

I'm new to this project but may use it heavily.  Therefore, I'm willing to volunteer myself for a fix.  But it seems like the kind of thing that should be discussed a little first.

verybadcat:
I just downloaded Math.Net numerics, and asked it to calculate SpecialFunctions.Gamma(5).  The correct answer is 24, but it returned 24.000000000000114.  

Can this be fixed?  (My proposed changes would fix this.  I don't really care how we fix it; if there is a better way to fix it, I would be fine with that.)

cdrnet:
I finally had a look at how the new algorithm performs numerically. It is indeed mostly 1-2 digits more accurate than the existing implementation and behaves very nicely around positive integers.

I noticed two potential problems so far:
* Accuracy drops up to 5 digits around 10e-5 (but is much better again at 10e-10).
* GammaLn no longer supports 0.0 and negative numbers (where Gamma is non-negative).

verybadcat:
I've got a fix for the accuracy issue, by simply using more terms of the power series and moving the lower breakpoint.  I'll post once I've had a look at the other issue.

verybadcat:
I've just posted a version that addresses both of your issues.  I'm returning NaN for GammaLn in cases where Gamma is negative.  But it's unclear if that's the right choice; one could instead return the real part.

https://github.com/mathnet/mathnet-numerics/pull/204/commits

PR #204, opened 2014-03-25
Gamma function

verybadcat:
I've found a C# version of code that appears to be used by e.g. Apple to compute the Gamma function, together with evidence that it's been around the web since the 1980s.  It needed some adaptation as the code appeared to be optimized for a smaller floating-point type than the 64 bits one typically sees today.  My adapted version produces exact answers at integers, which was a motivation for the change, and is within a factor of 10^-12 of the existing implementation elsewhere.

See the discussion here for background:

https://github.com/mathnet/mathnet-numerics/issues/196


verybadcat:
This fixes your issues about accuracy and also about GammaLn with negative inputs.  I'm choosing to return NaN for GammaLn in cases where the Gamma function is negative.  But it is arguable whether this is the right choice; one could return the real part of the log instead.

PR #265, opened 2014-11-14
Proof of concept for creating aspnet50 and aspnetcore50 based assemblies to include in the generated nupkeg

Tragetaschen:
This is just a (more or less) quick writeup to create the necessary infrastructure that creates the aspnet50 and aspnetcore50 assemblies for the nupkeg. This needs a KRuntime / VS2015 (https://github.com/aspnet/Home) to build.

I don't expect this to be merged. It just shows that providing those new targets through the nupkeg should not be too much of a hassle.

### DIFF
--- a/src/Numerics/Compatibility.cs
+++ b/src/Numerics/Compatibility.cs
@@ -1,4 +1,4 @@
-﻿#if PORTABLE
+﻿#if PORTABLE || ASPNETCORE50
 namespace MathNet.Numerics
 {
     using System;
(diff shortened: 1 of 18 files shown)


cdrnet:
Thanks (3 years later...)!

Closing, as it seems this is no longer needed now given the progress for .Net Standard support.

Tragetaschen:
I've been following the journey. Thanks for .NET standard support!

Issue #228, opened 2014-06-30
Exponential integral


whaddamoroon:
Incomplete Gamma function errors when a = 0. Would it be possible to add the exponential integral function to solve this?

PR #274, opened 2015-01-03
added correct handling for a=0 for regularized incomplete gamma function...

glyphard:
added correct handling for a=0 for regularized incomplete gamma function for upper and lower for issue #228
also added checksum test to verify that upper and lower incomplete regularized gamma function always sums to one for the same a and x.


cdrnet:
Thanks!

Issue #135, opened 2013-07-19
Feature request: numerical differentiation


MrCsabaToth:
A feature to calculate a value of a derivative of a function in a certain point.
http://stackoverflow.com/questions/17750770/how-can-i-calculate-first-and-second-derivative-of-a-function-in-c-sharp
Possible concept code:
http://stackoverflow.com/questions/373186/mathematical-derivation-with-c

PR #280, opened 2015-01-10
Numerical Differentiation Support

hsidky:
This is a pull request fulfilling #135 plus more features. 

I added support for numerical differentiation by creating two main components. A finite difference coefficients class and a numerical derivative class. As the name suggests, `FiniteDifferenceCoefficients` generates the coefficients necessary for numerical differentiation. For a given number of points, It pre-computes and stores all possible derivative orders and different centerings. This provides maximum flexibility (not restricted to forward, backward, center). See http://en.wikipedia.org/wiki/Finite_difference_coefficient .

Built on top of that is `NumericalDerivative` which uses the previous class to evaluate functions. It supports various step size types and I've included a few helper methods that give back derivative function handles. Basically all types of derivatives are possible. Equidistant points (data), variable order univariate and multivariate full and (mixed/nonmixed) partial derivatives respectively. My implementation of mixed partial derivatives is not the most efficient - there is propagation of error that's happening because it's recursive. I plan on implementing a 2D grid method to generate the coefficients in `FiniteDifferenceCoefficients` at a later date if this pull is merged. 

There are also two helper classes, `NumericalJacobian` and `NumericalHessian`. In the future, a method in `NumericalHessian` can be added that works with user provided first order derivative, rather than function. There are unit tests for all classes -  I believe I've got everything pretty covered, but there could be more for a few of them. I tried to keep the classes as lightweight as possible to facilitate merging, and thoroughly commented it. 

I know that there are pending non-linear optimization branches/pull requests. I think these numerical differentiation features would fit in perfectly for calculating Jacobians/Hessians for solvers. In fact, I have already written the groundwork for that, but there are already many different offerings. 

Hope this helps!

hsidky

### DIFF
--- a/src/Numerics/Differentiation/FiniteDifferenceCoefficients.cs
+++ b/src/Numerics/Differentiation/FiniteDifferenceCoefficients.cs
@@ -1,0 +1,140 @@
+﻿// <copyright file="FiniteDifferenceCoefficients.cs" company="Math.NET">
+// Math.NET Numerics, part of the Math.NET Project
+// http://numerics.mathdotnet.com
+// http://github.com/mathnet/mathnet-numerics
+// http://mathnetnumerics.codeplex.com
+//
+// Copyright (c) 2009-2015 Math.NET
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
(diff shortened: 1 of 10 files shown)


cdrnet:
Excellent, thanks a lot! Numerical differentiation has been on our list for a very long time.

MrCsabaToth:
Fantastic!

PR #282, opened 2015-01-14
Optimization: initial commit of candidate API and Newton implementation

scottstephens:
This is the first step of the plan to integrate the various optimization branches people have created, as discussed in https://github.com/mathnet/mathnet-numerics/pull/173. 

This should be pulled into its own branch in the mathnet/mathnet-numerics repository, not into mathnet/mathnet-numerics master.


cdrnet:
Done, thanks!

Issue #285, opened 2015-02-13
MKL provider should handle return codes of functions


tibel:
e.g. `SingularValueDecomposition()` should throw `NonConvergenceException` when `d_svd_factor()` return value is greater zero (https://software.intel.com/en-us/node/521150)

cdrnet:
Good point, yes, we should do that.

PR #290, opened 2015-02-18
Native Tweaks

cuda:
Simplified the LAPACK code by replacing the explicit function pointers definitions with template type deduction. It worked well, since it was able to deduce the correct values for complex eigen factorization - so the provider now supports it. Hopefully, this will simplify adding other providers. I'll try to add an Accelerate Framework/veclib provider and see if we can generalize the code a bit more.

I also added the non-convergence check on the native return value for SVD and eigen.


cdrnet:
Thanks, Marcus!

Related: #285

tibel:
Great, thanks

PR #295, opened 2015-03-09
Minkowski Distance

EMostafaAli:
Adding Minkowski Distance to the histogram class, it can be used to compare two histograms


cdrnet:
Thanks!

I wonder whether we can make it such that any of the [distance functions](http://numerics.mathdotnet.com/Distance.html) can be used here?

EMostafaAli:
Should I use delegates to accept any distance function? If yes, I believe there will be two overloads because Distance functions signature differ a little bit

EMostafaAli:
I believe this can be merged now in the code, it uses delegates to call distance functions

PR #296, opened 2015-03-10
Detect processor architecture and process bitness for loading native providers

kjbartel:
Added NativeProviderLoader, a helper class for loading native provider libraries depending on the processor architecture and the process bitness. MklLinearAlgebraProvider modified to call the helper class in InitializeVerify.

Not sure exactly how to automate building AnyCPU build though as both x86 and x64 builds need to be built first and then the files copied into subdirectories (x86 and amd64) of the AnyCPU output directory.

New nuget package for for combined x86 and x64 MKL will be needed. Couldn't find the nuspec though.

### DIFF
--- a/src/Numerics/Numerics.csproj
+++ b/src/Numerics/Numerics.csproj
@@ -192,6 +192,7 @@
     <Compile Include="LinearAlgebra\Vector.BCL.cs" />
     <Compile Include="LinearAlgebra\Vector.Operators.cs" />
     <Compile Include="Exceptions.cs" />
+    <Compile Include="Providers\NativeProviderLoader.cs" />
     <Compile Include="Random\SystemRandomSource.cs" />
     <Compile Include="Random\RandomSeed.cs" />
     <Compile Include="RootFinding\Broyden.cs" />
--- a/src/Numerics/Providers/LinearAlgebra/Mkl/MklLinearAlgebraProvider.cs
+++ b/src/Numerics/Providers/LinearAlgebra/Mkl/MklLinearAlgebraProvider.cs
@@ -134,11 +134,12 @@ namespace MathNet.Numerics.Providers.LinearAlgebra.Mkl
         /// </summary>
         public override void InitializeVerify()
         {
-            // TODO: Choose x86 or x64 based on Environment.Is64BitProcess
-
             int a, b, linearAlgebra;
             try
             {
+                // Load the native library
+                NativeProviderLoader.LoadNativeLibrary(SafeNativeMethods.DllName);
+
                 a = SafeNativeMethods.query_capability(0);
                 b = SafeNativeMethods.query_capability(1);
 
--- a/src/Numerics/Providers/LinearAlgebra/Mkl/SafeNativeMethods.cs
+++ b/src/Numerics/Providers/LinearAlgebra/Mkl/SafeNativeMethods.cs
@@ -46,322 +46,323 @@ namespace MathNet.Numerics.Providers.LinearAlgebra.Mkl
         /// <summary>
         /// Name of the native DLL.
         /// </summary>
-        const string DllName = "MathNet.Numerics.MKL.dll";
+        const string _DllName = "MathNet.Numerics.MKL.dll";
+        internal static string DllName { get { return _DllName; } }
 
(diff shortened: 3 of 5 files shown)


cuda:
This seems to be windows specific, since you are using loadlibrary from kernel32. Maybe this should be extended to work with dlopen for linux and osx, or add an OS check and fall back the standard p/invoke library resolution for non-windows platform. Also, maybe this should should be optional since users might want store the native libraries in one place and use envirorment variables for the p/invoke to find them (the native libraries are quite large).

kjbartel:
Yes you're right I didn't really think about Unix. I've added an OS check and different loaders for Windows and Unix however I'm not sure how to detect the cpu architecture on Unix. I did find [this page](http://mono.1490590.n4.nabble.com/Can-You-Programmatically-Determining-Architecture-x86-x64-IA64-td3858106.html) which suggests `uname()` on Unix but it didn't sound very promising.

cuda:
> I'm not sure how to detect the cpu architecture on Unix.

Me either. Googing around, I coudn't a decent solution.

cdrnet:
Right now we only support one of the 32 bit (x86) and one of the 64 bit families (amd64/x64). Then, in either case, we need to try them out by calling `InitializeVerify` (which by definition must fail on the wrong platform). Therefore, once we support more platforms for one of the providers (e.g. IA64) in the worst case we could try all candidates, with an educated guess on which one to try first.

Expected behavior:
* InitializeVerify must fail/throw on a platform which is not supported by a specific provider implementation
* Control.ConfigureAuto() must never fail/throw but may fall back to the managed provider.
* Control.UseNativeMKL() must fail/throw if it cannot load an initialize a native MKL provider implementation.

Hence, since our MKL provider does not support e.g. IA64, on this platform Control.ConfigureAuto must fall back to the managed provider and UseNativeMKL must fail. Same for ARM.

cdrnet:
@cuda regarding the environment variable I assume you're referring to the Path variable? It seems we'd need to probe all the paths in this variable ourselves to make this work with such an automatic solution?

Related: http://christoph.ruegg.name/blog/loading-native-dlls-in-fsharp-interactive.html

cuda:
>regarding the environment variable I assume you're referring to the Path variable? 

Chris, I was thinking more of LD_LIBRARY_PATH for Linux and DYLD_LIBRARY_PATH for OS X.

> It seems we'd need to probe all the paths in this variable ourselves to make this work with such an automatic solution?

I was suggesting to optionally not automatically load the libraries but let P/Invoke locate them. How it does so is well defined. I'm sure you've seen http://www.mono-project.com/docs/advanced/pinvoke/ .

cdrnet:
Ah, thanks!

kjbartel:
If an architecture specific subfolder isn't found then it will try to load the library without specifying a path resulting in normal searching behaviour. According to [dlopen man page](http://linux.die.net/man/3/dlopen) it will use LD_LIBRARY_PATH when searching. Likewise [LoadLibrary](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684175%28v=vs.85%29.aspx) will also search the PATH.
If that fails to locate and load the library then the function will return false and won't throw an exception. Instead of trying to use the searching within dlopen and LoadLibrary we could return false straight away in the case that the library isn't found within an architecture specific folder. Either way, if a library hasn't been loaded then normal P/Invoke will be used with normal exceptions being thrown etc. when any of the NativeNativeMethods functions are called.
Regarding ia64 and ARM detection support, I think that although currently MKL only supports x86 and amd64 there's no reason to think that in the future there won't other native providers which support other architectures. Indeed I've started working on getting OpenBLAS to work which supports ARM.

cuda:
>If an architecture specific subfolder isn't found then it will try to load the library without specifying a path resulting in normal searching behaviour.

That works me, thanks!

cdrnet:
Thanks again - this has been on my wish list for quite a while!

Next step will be to add a combined NuGet package (and while we're at it, finally add a props file so they get copied to the output directory automatically)

cdrnet:
Looks like I was a bit too fast with merging, the changes broke the portable builds (unfortunately not covered by CI) - will try to fix.

cdrnet:
No matter, fix is trivial.

kjbartel:
Ok. That looks like it should be be good now then. Thanks for that and glad I could help out.

PR #297, opened 2015-03-15
OpenBLAS native provider implementation

kjbartel:
I've only tested in Windows so far as I haven't had a chance to set up a linux vm and build OpenBLAS. I won't be able to test OSX, Android or iOS as I don't have a Xamarin license or a mac.

OpenBLAS should also work on ARM and other architectures but I'm not really sure about how to build OpenBLAS for mobile devices (iOS, android, WP8, WinRT) particularly as OpenBLAS is built with gcc and doesn't support MS vc++ or Intel compilers. For a start though it should work with x86, x64 and ia64 on Windows, Linux and OSX.


cdrnet:
Excellent, thanks a lot for moving forward with this!

cuda:
I was wondering about including the OpenBLAS Windows libraries in the repo. Is the idea that user's will build their own native wrapper? If so, then it makes sense since building OpenBLAS on Windows on isn't trival (at least for me).  I don't think we'd need to the same for Linux or OSX. 

Curious, were you able to link against libopenblas.a in VC++ (to remove the need of distributing libopenblas.dll)?  I wasn't able to. thanks

kjbartel:
Yes I wasn't sure whether to include the windows libraries in the repo or not but as I saw that nuget packages were included I figured you followed the notion of "source configuration management" and included libraries. Either way I'm not really fussed.

Regarding linking to libopenblas.a I believe it isn't possible due to being built with gcc and mingw. See this [OpenBLAS page](https://github.com/xianyi/OpenBLAS/wiki/How-to-use-OpenBLAS-in-Microsoft-Visual-Studio) discussing use of OpenBLAS with VC++. I was tempted to try and build OpenBLAS with VC++ 2013 (changing to use cmake) given that the latest compiler has better C++11 and C99 support. Either that or the Intel compilers. VC++ I think would be the only way to get a Windows ARM build (for WinRT / WP8) but honestly there's probably so little demand for it it's not worth the effort. And given my knowledge of C++ is a bit limited I didn't look into it further.

cuda:
> Regarding linking to libopenblas.a I believe it isn't possible due to being built with gcc and mingw.

Do you know if  the wrapper DLL can be built with mingw, statically linking libopenblas.a, and still be P/Invoke'd? thanks

kjbartel:
> Do you know if the wrapper DLL can be built with mingw, statically linking libopenblas.a, and still be P/Invoke'd?

Probably yes. But it would then require the setup of MinGW-w64 (x86 and x64 versions) just for building the wrapper as opposed to only requiring VS currently. Is there any particular benefit you see in static linking as opposed to dynamic linking? As OpenBLAS has already been manually optimized the only difference I can think of is the one extra dll with dynamic. [Static vs dynamic](http://stackoverflow.com/a/1993407/1730559).

cuda:
>But it would then require the setup of MinGW-w64 (x86 and x64 versions)

Right, but you have those enviroments already setup to build OpenBLAS. I was just thinking if the libraries are already built, should we just go the next step and provide the wrapper DLL instead of the libraries. Then statically link to reduce the number of files we need to distribute. We should be able to get rid of libopenblas, libgfortran, and libquadmath. That would just leave the wrapper DLL and the exception handling DLL. Let me tinker with something.....

cuda:
>to reduce the number of files we need to distribute.

But of course, that might not even be an issue.

cuda:
I tried to build the wrapper on Ubuntu 14.4 but gcc 4.8 doesn't like the &(alpha.real), &(x->real) syntax.
Changing it to (float*)&alpha, (float*)x and changing complex_qr_solve_factored to:
https://gist.github.com/cuda/a8e3de990986c2a57e58
seemed to work.

kjbartel:
> Right, but you have those enviroments already setup to build OpenBLAS.

Actually, I don't. I just used the pre-built binaries provided by OpenBLAS on [sourceforge](http://sourceforge.net/projects/openblas/files/). And I was thinking also for anyone who got the Math.Net source from github. I actually did install MinGW-w64 and MSys and tried to build OpenBLAS with no success.

> gcc 4.8 doesn't like the &(alpha.real), &(x->real) syntax

I'm guessing because they're using different complex definitions. I was trying to avoid unnecessary casts.

> Do you know if the wrapper DLL can be built with mingw, statically linking libopenblas.a, and still be P/Invoke'd?

I've been thinking about this more and if it can be done then there's really no reason to have a wrapper dll at all. You might as well just P/Invoke directly to libopenblas.dll and move the code currently in the wrapper into C#. You'd still have exactly the same number of files though. The only reason I could see to actually have a wrapper would be if it were mixed mode managed / native and then not use P/Invoke but instead load the library with normal `Assembly.LoadFrom`.

> to reduce the number of files we need to distribute.

I personally don't feel it's an issue. As long as a nuget package can be put together which has all the required files set as project content.

If you really wanted to reduce the number of files then you could put everything as resources within an assembly, but native libraries [need to be written to a folder before loading](http://stackoverflow.com/questions/666799/embedding-unmanaged-dll-into-a-managed-c-sharp-dll). Considering the main purpose of using a native provided is to increase performance I don't think this is a good option just to reduce file numbers.

I'm still curious if there'd be another way to build OpenBLAS on Windows to get rid of MinGW. Intel compilers or some combination of LLVM / clang / dragonegg both of which should result in libs which can be statically linked in VC++. But it'd probably require us to fork OpenBLAS and clean up the build with cmake. That'd be quite a bit more work than what's in this pull-request.

cuda:
>Actually, I don't. I just used the pre-built binaries provided by OpenBLAS on sourceforge. And I was >thinking also for anyone who got the Math.Net source from github. I actually did install MinGW-w64 and >MSys and tried to build OpenBLAS with no success.

Ah, OK. That answers the question about static linking then :)

> I'm guessing because they're using different complex definitions. I was trying to avoid unnecessary casts.

Didn't think about that. I'll take a look, thanks! As for the casts, there shouldn't be any overhead - pointer casting is almost costless (as I understand it). It was suggested on the OpenBLAS users group. But I'll tinker with the complex definitions this weekend.

> If you really wanted to reduce the number of files then you could put everything as resources within an assembly, 

I personally have no interest in going that route. If static linking could reduce the number and size of the files, great.  Again, it probably isn't an issue. Just my OCD trying to minimize everything.

>You might as well just P/Invoke directly to libopenblas.dll and move the code currently in the wrapper into C#.

You could do that. The original idea was to have a single interface that all native libraries would provide and to minimize the number of P/Invoke calls (just provide what the native library the required inputs,  let it do its business, and hand back the result).  We had the gradnose idea of supporting CUDA, DirectCompute, OpenCL, and other libraries which would have required more P/Invoke calls than CBLAS and LAPACK. Though, now there is cuBLAS, MAGMA, and others, so maybe this needs to rethought.

kjbartel:
Currently I haven't added native implementations of the vector functions present in MKL but not provided by OpenBLAS. Should be fairly simple to add native versions by either calling matrix functions with single column or row matrices or calling other vector functions which are present. I'm not so sure there'd be any real advantage in doing this over the current managed parallel implementations for the vector functions though because of the need to create matrices from the vectors, resulting in a lot more copying of data. Haven't tested it out yet but I thought I'd get your thoughts.

> We had the grandiose idea of supporting CUDA, DirectCompute, OpenCL, and other libraries which would have required more P/Invoke calls than CBLAS and LAPACK. Though, now there is cuBLAS, MAGMA, and others, so maybe this needs to be rethought.

My main reason for starting with OpenBLAS was due to it being open source and having very good performance compared to other libraries. The other library / technology I was interested in was OpenCL but it looks to be a bit more difficult to implement. OpenCL would be good as it can run on both CPUs and GPUs from all vendors and OS platforms. Cuda has similar performance to OpenCL but due to it only being on nVidia GPUs it didn't really interest me.

The main downside to OpenBLAS is that it currently can only be built on Windows using mingw / gcc so you end up using multiple runtime libraries. I did actually manage to build OpenBLAS using clang and gfortran on Windows but some unit tests would throw AccessViolationException using that dll so that's not going to work. I couldn't find a copy of mingw/gcc with working dragonegg plugins to try and use that instead, particularly for fortran. The next version of Visual Studio and the C++ compiler apparently has improved quite a lot and clang is also going to be supported in VS for building native libraries for Android and iOS. Still doesn't mean that OpenBLAS could be built using VS though.

cuda:
> I'm guessing because they're using different complex definitions.

yep, it is using the C99 complex types. Still working on it...

>I'm not so sure there'd be any real advantage in doing this over the current managed parallel implementations for the vector functions though because of the need to create matrices from the vectors, resulting in a lot more copying of data. 

Both the dense matrix and vector storage classes expose their underlying arrays, so there is very little copying if you are working directly on them.  The only advantage would be if the the native side takes advantage of SIMD operations (the complier's auto vectorization might work). If not, then yes, it wouldn't be worth the effort. 

The nice thing about getting a SIMD version of the vector operations is that it can then be used with other providers that don't support vector functions, such as ATLAS.

kjbartel:
> yep, it is using the C99 complex types. Still working on it...

I wouldn't worry about it. I've already changed to using casts as you suggested earlier.

> Both the dense matrix and vector storage classes expose their underlying arrays, so there is very little copying if you are working directly on them. The only advantage would be if the the native side takes advantage of SIMD operations (the complier's auto vectorization might work). If not, then yes, it wouldn't be worth the effort.

Doing a bit of performance testing it seems that using the managed implementation is faster than using blas functions such as axpy and sbmv. Simply implementing the vector functions in C++ using for loops which were then auto vectorized by the compiler gave performance almost the same as the managed implementation but was still slower. And using a parallel for with OpenMP gave slightly slower results again than the auto-vectorized version. So...  looks like there's no advantage as the managed parallel for implementation gives the best performance, at least on my computer and for the arrays I was testing.

cuda:
>I wouldn't worry about it. I've already changed to using casts as you suggested earlier.

When I was testing further, I ran into some segfaults on Linux. I want to make sure the C99 complex data types are not the problem.

>Doing a bit of performance testing it seems that using the managed implementation is faster

Interesting. I would not have expected that. I'll benchmark the the MKL vector fuctions and see if the same holds.

redknightlois:
For small and not so small sizes even using SupressUnmanagedCodeSecurity is faster. Last weekend I have been doing some perf tests for a different reasons on simple array operations, and noticed that the JIT compiler is actually quite smart regarding loop unrolling for SAXPY like operations. So unless there is real vectorization (SIMD code) you will see not much (if any) improvement over a native C# array based C#. Even unsafe pointers is not good on certain setups. 

Those are some simple measurements.
https://gist.github.com/anonymous/a5e876e406f5b1ac9608

kjbartel:
> When I was testing further, I ran into some segfaults on Linux. 

If they are happening in lapack and only in complex functions then try putting a `#ifdef` around the custom complex definitions at the top of the lapack.cpp file:

```C++
#ifdef OPENBLAS_COMPLEX_STRUCT
#include "complex.h"
#define LAPACK_COMPLEX_CUSTOM
#define lapack_complex_float complex<float>
#define lapack_complex_double complex<double>
#endif
```
I added the complex definition so that the code could be kept pretty much exactly the same as the MKL code. MKL uses `std::complex<>` though instead of the struct (basically the same though). I thought I ran into some problems using `std::complex<>` but testing again now it seems that also works. So you could also change to:
 ```C++
#include <complex>
#define LAPACK_COMPLEX_CUSTOM
#define lapack_complex_float std::complex<float>
#define lapack_complex_double std::complex<double>
```

cuda:
> When I was testing further, I ran into some segfaults on Linux.

Turned out to be a problem with my compile command.

cuda:
>Interesting. I would not have expected that. I'll benchmark the the MKL vector fuctions and see if the same holds.

Some quick tests comparing the pointwise add, subtract, divide, and multiply between the managed and MKL provider (which uses the MKL vector functions) I'm seeing the the MKL functions about 4x faster. If we could do a SIMD implementation (I guess the auto vectorization isn't doing a very good job), it might be worth it.

kjbartel:
> Turned out to be a problem with my compile command. 

I had a couple of builds on Windows which had problems like that too.

> If we could do a SIMD implementation (I guess the auto vectorization isn't doing a very good job), it might be worth it.

Interesting. I originally got similar results for OpenBLAS but I realised afterwards that it was because I had a debugger attached. I did actually see that OpenBLAS added ?geadd functions in the latest release a few days ago which could be used for add and subtract and maybe get better performance. It will also depend a lot of the size of the arrays as small arrays should be faster single threaded.

Anyway the question is whether to put SIMD vector function implementations as part of this pull request or a separate one just for that. Other than the vector functions is there anything else which needs to be done?
* Change the version number so it's not the same as MKL.
* Add Linux build script.
* Nuget package.
* Include or not include pre-built OpenBLAS binaries.
* ARM build for android or iOS?

I won't be able to work on this for a few days though as I'm pretty busy this week.

cuda:
>Include or not include pre-built OpenBLAS binaries.

My two cents, don't include them. The library is trival to build on Linux and OS X, and as you mentioned, users can download the pre-bulit version for Windows. Only a handful of people will build the wrapper if it is available via nuget.

cdrnet:
I plan to have a go at this this weekend, decide how to deal with the binaries, and finally merge it in.

kjbartel:
Ok. I haven't had much of a chance to look further at this recently. Had been meaning to do a bit more benchmarking of vector functions. A more efficient implementation could always be added at a later date though.

cdrnet:
I've pulled the OpenBLAS provider to master - thanks again!

I did not do a direct merge but filtered out the included OpenBLAS lib. The build will look for it at the `OpenBLASPath` environment variable, or fall back to `repo/../libs/OpenBLAS`.

kjbartel:
Ok thanks for that. I've gotten the latest source and it all looks good.

> It would be good if we could make libs/OpenBLAS available somewhere anyway - and maybe even pull it in locally using paket.

Yes I was thinking the same thing.

PR #298, opened 2015-03-19
Native provider loader fixes

kjbartel:
Couple of fixes to problems I found with the `NativeProviderLoader`.
* Couldn't find / load library dependencies in the architecture directory -> changed to `LoadLibraryEx`
* MSTest / Xunit would fail to locate the native library due to copying assemblies to Temp -> look in `AppDomain.Current.BaseDirectory` as well as `Assembly.GetExecutingAssembly().Location`.

### DIFF
--- a/MathNet.Numerics.NativeProviders.sln
+++ b/MathNet.Numerics.NativeProviders.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.30501.0
+VisualStudioVersion = 12.0.31101.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Common", "Common", "{5A0892FF-82CE-40FC-BCE1-73810C615F52}"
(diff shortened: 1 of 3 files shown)


kjbartel:
Question:
Do you prefer 'x64' or 'amd64' for the architecture directory name? Currently builds for the MKL wrapper go to an 'x64' directory so could change to 'amd64' or could change the NativeProviderLoader to use 'x64'. I'm not sure if there's a standard folder name as I've seen programs use both (even different Microsoft programs).

Either way I think the UnitTest-MKL project should be changed to only build "And CPU" builds and the output changed to "out\MKL\Windows\" and hence will load the correct library from it's sub-directories depending on whether the tests are run as x86 or x64.

cdrnet:
I prefer `x64`, `x86`, `ia64`, etc. where `x64` refers to both amd64 and intel64 (EM64T) as common on Windows, even though they're equivalent. I believe in Linux this is usually referred to as `amd64`, others seems to use `x86_64`. At least for Intel MKL I think it would be somewhat confusing to call it amd64.

I agree on building the test project for AnyCPU - should make things simpler and may also serve as practical poc+test for the native loader.

kjbartel:
Ok changes made.

PR #303, opened 2015-04-10
Add (restore?) the missing SparseVector.AbsoluteMaximumIndex() override implementations

MattHeffron:
The _type-specific_ SparseVector implementations **DID NOT** have AbsoluteMaximumIndex overrides although they **DO** have AbsoluteMinimumIndex overrides.

From the erroneous comment headers for MaximumIndex (which immediately follows AbsoluteMinimumIndex in the Double/ and Single/ files), it _appears_ that the implementations were all removed accidentally.


cdrnet:
Great, thanks!

PR #305, opened 2015-04-21
chisquared inverse distribution

logophobia:
Utility method to get the quantiles for the chisquared distribution. This can be done through the gamma distribution, but this is a bit nicer.

Also added a unit test


cdrnet:
Great, thanks!

PR #306, opened 2015-04-26
MathNet.Numerics.CUDA native provider via integration with cuBLAS

borfudin:
I have added a new Native Provider a la the existing MKL provider for NVidia graphics cards by utilizing the cuBLAS and cuSolver APIs in the CUDA Toolkit:

https://developer.nvidia.com/cublas
https://developer.nvidia.com/cusolver

This entailed the following additions:
- MathNet.Numerics.CUDA.dll C++ library for PInvoke purposes
- MathNet.Numerics.Providers.LinearAlgebra.Cuda namespace with requisite classes
- UnitTests-CUDA project, set up to run all tests using the CUDA native provider

This is very similar to the MKL native provider with a few exceptions due to limitations of the toolkit and/or appropriateness:
- There is no "vml" style vector manipulation as the cost of transferring the data to the device would in many cases outweigh any benefit that might be gained and are handled by the managed provider instead.  That said, "?axpy" style math still goes to the device.
- The cuSolver API does not currently have robust support for QR operations due to some key missing functions (like ?ormqr and ?gels) and so all of those functions use the managed provider.
- SVD is full supported, but again due to limitations in the API operations which used the "eigen_factor" functions instead use the managed provider.

I hope this proves useful.


borfudin:
I should mention this has a build dependency on the NVidia CUDA toolkit:

https://developer.nvidia.com/cuda-toolkit

The project file uses the environment variables and therefore should build on any machine which has the toolkit installed, but I've only tested this on a Windows machine using VS 2013 so far.

cdrnet:
Thanks a lot, this will indeed be very useful!
I can take care of integrating this into the FAKE build automation.

Do you know whether Nvidia's licensing allows us to publish the compiled provider binaries including cublas64_70.dll as NuGet packages (like we do for Intel MKL) and whether any user is then free to redistribute this as part of their application without requiring any license themselves (unlike MKL)?

I did a few quick tests on my system (with a rather old Quadra 4000), some operations worked very well (e.g. LU decomposition) but I run into some issues with SVD and matrix multiplication, will need to investigate further.

borfudin:
Hey-
 
Glad you like it!  Some answers:
 
1) I am not a lawyer, but my reading of NVidia's license is that they are giving away the software freely: http://docs.nvidia.com/cuda/eula/index.html#nvidia-cuda-toolkit-license-agreement.  It doesn't particularly surprise me if that is the case, as it drives sales of their cards.  The pertinent bit is here:
 
"Licensee may transfer, redistribute, and sublicense certain files of                                    the Redistributable SOFTWARE, as defined in Attachment A of this Agreement, provided, however, that                                    (a) the Redistributable SOFTWARE shall be distributed solely in binary                                    form to Licensee's licensees ("Customers") only as a component of                                    Licensee's own software products (each, a "Licensee Application"); (b)                                    Licensee shall design the Licensee Application such that the                                    Redistributable SOFTWARE files are installed only in a private                                    (non-shared) directory location that is used only by the Licensee                                    Application; (c) Licensee shall obtain each Customer's written or                                    clickwrap agreement to the license terms under a written, legally                                    enforceable agreement that has the effect of protecting the SOFTWARE                                    and the rights of NVIDIA under terms no less restrictive than this                                    Agreement. "


I _think_ MIT/X11 satisfies (c), but again not an expert here.
 
2) It is not entirely surprising that you may be having issues with older cards, as functionality in the toolkit is hardware and driver dependent.  That said, matrix multiplication should work everywhere and the code passed all the matrix math tests on my local.  Happy to continue to work with you on any bugs you find.
 
Hope this helps!
 
Regards,
 
Matthew
 
Date: Tue, 28 Apr 2015 22:15:30 -0700

cdrnet:
Is there an easy way to detect what is supported by the actual GPU, so we can fall back to managed (or to another provider) automatically?

borfudin:
Yes, the library returns response codes that I'm currently handling in a crude way and you could use those to fail over to the managed code.  That said, I think it may be better to handle the response codes differently and throw an informative exception in those circumstances so as to indicate to the user that using the CUDA native provider with that hardware may not be the optimal choice.  Otherwise it might create some weird "debugging" scenarios where a user is trying to figure out why something is so slow, thinks it is the graphics card, but it is in fact because the system has been silently failing over to the managed provider.

cdrnet:
If we can get clean exceptions that would be an option. Currently on my system svd just hangs and on multiplication I get an exception, but only after the first call.

Nvidia seems to use compute capability levels to deal with compatibility. Apparently my card has cc 2.0 only. Maybe this would also be a way to deal with this. E.g. we could define a minimum required CC level below which the provider declares itself as incompatible and fails to initialize, and then maybe have a subset of functionality which is only used on certain CC levels, below those it could fall back silently.

cdrnet:
I've just merged, let's discuss in new issues instead.

borfudin:
Sure thing.  I think requiring a minimum level is probably the best approach, and would be fairly easy to do on initialization of the native provider.

Issue #283, opened 2015-01-14
Merge Optimization Efforts


scottstephens:
As discussed in https://github.com/mathnet/mathnet-numerics/pull/173, we've got a few independent efforts to add optimization to mathnet-numerics, and we would like to merge these efforts, and eventually get the code into the mathnet-numerics master.

I'm starting this issue so we have a central place to discuss these efforts, since it will likely involve multiple pull requests.

The first thing we need to do is get the various contributors to settle on an API. To that end, I've created a pull request (https://github.com/mathnet/mathnet-numerics/pull/282) with the start of some experimental code. Right now, it's basically got the API, and an implementation of a Newton unconstrained minimizer. It's branched from the current mathnet-numerics master.

Suggested next steps:
(1) @cdrnet  should pull my branch into it's own mathnet branch
(2) @joemoorhouse should add an algorithm that calls out to native code using the candidate API to see how he likes working with it/how native code fits in.
(3) Any interested parties, including at a minimum me, @joemoorhouse, and @cdrnet should discuss the API
(4) Once the API is settled, we can add additional algorithms.

cdrnet:
(1) Done: [unified_optimization branch](https://github.com/mathnet/mathnet-numerics/tree/unified_optimization).

joemoorhouse:
Hi Scott, Just had a look at the API, and personally I like the concept. I think it is nice to have an interface in which the Minimizer asks for the function value, gradient or Hessian (and in turns says when the point at which these are calculated has changed). I think that is going to give enough flexibility where calculations of these share computationally intensive components.

I have one idea (which might or might not be a good one! :)): I have been playing around with the interfaces and wonder whether it is possible to simplify (arguably) somewhat, because I see that the IEvaluation is already responsible for calculating and storing the function value, gradient and Hessian. I wonder if we can combine IObjectiveFunction and IEvaluation and the objects that implement these into one. The single interface would be:

    public interface IObjectiveFunction
    {
        Vector<double> Point { get; set; }
        double Value { get; }
        Vector<double> Gradient { get; }
        Matrix<double> Hessian { get; }
    }

I think that EvaluationStatus can (even should?) be part of the internals. The setter of the Point could be used to indicate when this has been changed and be used to reset EvaluationStatus. 

Anyway, I have put an illustration of this here (and can flesh out more if required). 

[...]

PR #308, opened 2015-05-03
Implementation of Broyden–Fletcher–Goldfarb–Shanno algorithm

braddodson:
I needed this for numerical optimization so it seemed worth making it a part of Math.net numerics.


cdrnet:
Great, thanks a lot!

I've merged b858289 into the unified_optimization branch in order to integrate it with the ongoing effort of #283.

PR #309, opened 2015-05-06
Small value histogram

jneedhamspkt:
When using Histogram on a set of small doubles, I observed that the buckets were not sorting properly, which resulted in GetBucketIndexOf() throwing ArgumentExceptions before my Histogram was even constructed. 

I believe that the root issue here was that CompareTo() is using AlmostEqual() to compare bucket locations, which limits the resolution of the buckets to Precision.DefaultDoubleAccuracy

This request adds a set of unit test for small values, and safely moves the CompareTo function to .Equals().


cdrnet:
Looks good to me otherwise, thanks!

jneedhamspkt:
No problem. Happy to contribute!

Issue #140, opened 2013-10-02
Signed Edition issue with MathNet.Numerics.FSharp, MathNet.Numerics.Data and MathNet.Numerics.Native


tibel:
At the moment it is not possible to use MathNet.Numerics.FSharp or MathNet.Numerics.Data from a strong-named assembly. Same is true for using the native providers as the nuget package depends on the unsigned version.

I think there should only be one version of the library and this should be the signed version. Nuget now supports adding version redirect rules to App.Config to deal with versioning issues.

So rethink your decision with signed and unsigned edition, please.

cdrnet:
Picking this up again. There is a weak and unverified reason regarding MKL licensing when offering the MKL binaries independently, as we're only allowed to ship them as part of Math.NET Numerics (to my understanding). But given that it is actually statically compiled into the MathNet.Numerics.MKL.dll this is actually the case, so it should be possible to drop the package dependency. This was blocking not only in the -signed scenario but was also inconvenient when trying to control the native provider of choice from application level, see #312.

PR #312, opened 2015-05-13
Change Nuget targets to use 'Content' rather than 'Copy' and 'Delete'

kjbartel:
The 'Copy' and 'Delete' tasks will only work for a project which is directly executed (an executable). It won't work for any projects which reference the project which includes one of the MathNet nuget packages (such as a unit test project). Installer projects such as WiX or VS Setup projects similarly won't find the required library files unless they are 'Content'.

### DIFF
--- a/build/MathNet.Numerics.CUDA.Win.targets
+++ b/build/MathNet.Numerics.CUDA.Win.targets
@@ -8,67 +8,20 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!--
   ******************************************************************************
-  **                    MathNet Interop Library Build Items                   **
+  **                    MathNet Interop Library Files                         **
   ******************************************************************************
   -->
 
-  <ItemGroup>
-    <MathNetInteropFiles Condition="'$(MSBuildThisFileDirectory)' != '' And
-                                   HasTrailingSlash('$(MSBuildThisFileDirectory)')"
-                        Include="$(MSBuildThisFileDirectory)**\*.dll" />
-  </ItemGroup>
-
-  <!--
-  ******************************************************************************
-  **                   MathNet Interop Library Build Targets                  **
-  ******************************************************************************
-  -->
-
-  <Target Name="CopyMathNetInteropFiles"
-          Condition="'$(CopyMathNetInteropFiles)' != 'false' And
-                     '$(OutputPath)' != '' And
-                     HasTrailingSlash('$(OutputPath)') And
-                     Exists('$(OutputPath)')"
-          Inputs="@(MathNetInteropFiles)"
-          Outputs="@(MathNetInteropFiles -> '$(OutputPath)%(RecursiveDir)%(Filename)%(Extension)')">
+  <ItemGroup Condition="'$(MSBuildThisFileDirectory)' != '' And
+                         HasTrailingSlash('$(MSBuildThisFileDirectory)')">
     <!--
         NOTE: Copy "MathNet.Numerics.CUDA.dll" and all related files, for every
               architecture that we support, to the build output directory.
     -->
-    <Copy SourceFiles="@(MathNetInteropFiles)"
-          DestinationFiles="@(MathNetInteropFiles -> '$(OutputPath)%(RecursiveDir)%(Filename)%(Extension)')" />
-  </Target>
-
-  <!--
-  ******************************************************************************
-  -->
-
-  <Target Name="CleanMathNetInteropFiles"
-          Condition="'$(CleanMathNetInteropFiles)' != 'false' And
-                     '$(OutputPath)' != '' And
-                     HasTrailingSlash('$(OutputPath)') And
-                     Exists('$(OutputPath)')">
-    <!--
-        NOTE: Delete "MathNet.Numerics.CUDA.dll" and all related files, for every
-              architecture that we support, from the build output directory.
-    -->
-    <Delete Files="@(MathNetInteropFiles -> '$(OutputPath)%(RecursiveDir)%(Filename)%(Extension)')" />
-  </Target>
-
-  <!--
-  ******************************************************************************
-  **                 MathNet Interop Library Build Properties                 **
-  ******************************************************************************
-  -->
-
-  <PropertyGroup>
-    <BuildDependsOn>
-      $(BuildDependsOn);
-      CopyMathNetInteropFiles;
-    </BuildDependsOn>
-    <CleanDependsOn>
-      $(CleanDependsOn);
-      CleanMathNetInteropFiles;
-    </CleanDependsOn>
-  </PropertyGroup>
+    <MathNetInteropFiles Include="$(MSBuildThisFileDirectory)**\*.dll" />
+    <Content Include="@(MathNetInteropFiles)">
+      <Link>%(RecursiveDir)%(FileName)%(Extension)</Link>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
 </Project>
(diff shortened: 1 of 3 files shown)


tibel:
I add the native provider to the executable project and NOT the project containing the actual math algorithms.

I would prefer if the native provider nuget packages don't require a reference to MathNet package:
- when using MathNet.Numerics-Signed
- when the executable project does not have a reference to MathNet.Numerics

kjbartel:
> I add the native provider to the executable project and NOT to the project containing the actual math algorithms.

This is not a very good solution when you have multiple projects referencing the project using MathNet. For example, in the solution I'm working on I'd be required to add the native provider to two executable projects and a unit test project. I also have a VS setup project which knows nothing about the copied library files so it requires them to be added individually to the package. If the library files change in a later release of a provider nuget package then it would require manual modification again.

Changing to 'Content' I believe is the correct way to make sure the libraries are copied to where ever they are used and would only require the single reference to the nuget package from the project which actually uses the package.

tibel:
I run my tests on the build servers with:
- 32bit and 64bit
- with native provider and without

To be sure all possible combinations will work.

kjbartel:
> to be sure all possible combinations are working as expected.

Only if your unit test project also references the native provider nuget package.

kjbartel:
Another approach would be to not use targets files at all and instead have the library files under 'Content' instead of 'Build' in the nuget package. This has the problem that content files don't have `CopyToOutputDirectory` set to `Always` though. To fix that problem a powershell script would need to be run on install like [shown here](http://stackoverflow.com/questions/21143817/set-content-files-to-copy-local-always-in-a-nuget-package). Powershell scripts will only be run from within VS though so won't work when using nuget from the commandline and from other tools such as paket and build servers.

So, I'm still of the opinion that setting content in the targets file is the best solution.

cdrnet:
Conceptually I agree with @tibel that the native provider to use should be a choice of the application/host rather than a library in between. But in the end that's not a choice we can make for our users. And it would require #314.

NuGet's own Content folder behavior is exactly what we've been doing up to now and still do for now for the MKL provider in the packages with a CPU architecture suffix (just not in the new combined package). This worked quite well but means that there's always a manual step involved which I've always found very unsatisfying.

PowerShell in NuGet is a No-Go.

I quite like the automatic behavior with the new target file, but I understand it may not work well in all scenarios. Maybe there's a combined way?

@tibel re -signed, this is related to #140, I've picked up the discussion there again.

cdrnet:
If I understand it correctly, the target file proposed by this PR would actually remove the manual step as well - just with the drawback that the binary files will appear explicitly in the VS project. @kjbartel can you confirm this?

kjbartel:
> This worked quite well but means that there's always a manual step involved which I've always found very unsatisfying.

It is I agree. The benefits of using the normal nuget content folder though are that the added files are clearly visible in the project and are included by all other referencing projects.

>Installer projects such as WiX or VS Setup projects similarly won't find the required library files unless they are 'Content'.

I was mistaken that this solution would fix this problem. Seems for at least the VS setup project it will only pick up content files which are within the project file itself so put the content in a separate targets file doesn't solve the problem as I'd hoped.
  
> PowerShell in NuGet is a No-Go.

Any particular reason? I've been reading a few discussions regarding this and it seems the thinking of nuget developers is that installing packages should be done from within VS. It doesn't make sense to do it outside of VS as the project you add the package to needs to have code written to use the package. The argument falls down if you want to support non-VS or non-Windows development environments. Adding the script should support the majority or users I'd think. But, even if the script isn't run you just end up with the current situation requiring the manual step. So adding the script ends up only being a net positive.

I think I've convinced myself that the powershell script option with the normal NuGet Content folder is the best option. For me it seems to solve every problem I see:

* only need one reference to native provider package in same project as where MathNet package is referenced and used
* included files are visible in the project and not silently copied
* required library files are copied to all referencing projects
* setup projects detect all content files which need to be included

kjbartel:
> If I understand it correctly, the target file proposed by this PR would actually remove the manual step as well - just with the drawback that the binary files will appear explicitly in the VS project.

Yes it removes the manual step. However, the binary files do not get shown in the project. It is exactly the same as the current targets files in that regard.

See my previous comment though as I'm actually thinking now that the PR isn't the best solution.

kjbartel:
I've created a simple PS script for the MathNet.Numerics.MKL.Win package just to test. Seems to work ok.

```powershell
param($installPath, $toolsPath, $package, $project)

$x86 = $project.ProjectItems.Item("x86");
$x64 = $project.ProjectItems.Item("x64");
$x86.ProjectItems.Item("libiomp5md.dll").Properties.Item("CopyToOutputDirectory").Value = 1
$x86.ProjectItems.Item("MathNet.Numerics.MKL.dll").Properties.Item("CopyToOutputDirectory").Value = 1
$x64.ProjectItems.Item("libiomp5md.dll").Properties.Item("CopyToOutputDirectory").Value = 1
$x64.ProjectItems.Item("MathNet.Numerics.MKL.dll").Properties.Item("CopyToOutputDirectory").Value = 1
```

cdrnet:
I happen to disagree with the NuGet team in this regard (and others as well). In fact it's one of the reasons I use [Paket](http://fsprojects.github.io/Paket/) to maintain my NuGet packages and references. That's why you'll find no `packages.config` files in this very repo. In my opinion a package should be *declarative* and *neutral* to both IDE and platform.

However, native provider packages are a bit different as they are specific to a platform anyway, at least for now. So it is possible to deviate from these principles a bit here.

So the only thing the PowerShell script you suggest would do is to add the `<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>` tag? I assume this hybrid approach is not possible with a target or props file instead of a script?

cdrnet:
@tibel any thoughts on the PS script mentioned above? Would this work in your scenario as well?

tibel:
It should work I think.

This is my personal opinion only:
I don't like x86 and x64 folders in my projects.

cdrnet:
Yes, me neither.

kjbartel:
> So the only thing the PowerShell script you suggest would do is to add the <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory> tag?

Currently it's `<CopyToOutputDirectory>Always</CopyToOutputDirectory>`, but yes that's all it does.

> I assume this hybrid approach is not possible with a target or props file instead of a script?

I can't think of a way no.

cdrnet:
Untested though: maybe `<Visible>false</Visible>` could work to hide them

kjbartel:
> Untested though: maybe `<Visible>false</Visible>` could work to hide them

If I modify my project file manually then yes it works.

Modifying the PS script to make the modification is a bit more involved though as `Visible` is VS specific metadata and hence can't be set in the same way as the `CopyToOutputDirectory` property. Would need to instead use XML to find the element and add the `Visible` metadata using something like [here](http://stackoverflow.com/a/12292300/1730559) or [here](http://stackoverflow.com/a/21161498/1730559).

cdrnet:
I just experimented a bit with the version in this PR and it works very well for me even with project-references to a library where only the library has proper NuGet dependencies. Same with Unit Testing. Even Paket can deal with it properly (which skips PowerShell scripts).

The original target worked as well, but declaring the binaries as content like in the modified version is more declarative than explicitly copying them, so this and the fact that it is a bit simpler seems to be an improvement.

So all in all the VS installer project is the only case known so far where this wouldn't work out of the box in the *indirect* case. A known workaround is to add a direct package dependency to the installer (or its entry point project, I don't know much about VS installer projects), another one is to manually add Content links manually to the indirect project. This seems acceptable to me.

kjbartel:
> So all in all the VS installer project is the only case known so far where this wouldn't work out of the box

Yep. Installer projects are notoriously bad at finding dependencies. There are other nuget packages I use which have the same problem and I have to deal with anyway so it's no big deal. The project references is the main thing.

Thanks.

cdrnet:
Great, thanks again!

PR #319, opened 2015-05-27
Add a method to construct a beta PERT distribution

jbowtie:
A beta PERT derives the shape parameters from the min, max, and mode -- this is often used in risk analysis, disease management, and other domains where an expert forecast or assessment is required.

This PR adds a static PERT method to the Beta class to instantiate a beta distribution from the min, max, and mode parameters. Note that I named the mode parameter "likely" to better match other libraries but would be happy to rename.

### DIFF
--- a/src/Numerics/Distributions/Beta.cs
+++ b/src/Numerics/Distributions/Beta.cs
@@ -92,6 +92,48 @@ namespace MathNet.Numerics.Distributions
             _shapeB = b;
         }
 
+
+        /// <summary>
+        /// Create a Beta PERT distribution, used in risk analysis and other domains where an expert forecast
+        /// is used to construct an underlying beta distribution.
+        /// </summary>
+        /// <param name="min">The minimum value.</param>
+        /// <param name="max">The maximum value.</param>
+        /// <param name="likely">The most likely value (mode).</param>
+        /// <returns>The Beta distribution derived from the PERT parameters.</returns>
+        public static Beta PERT(double min, double max, double likely)
+        {
+            if( min > max || likely > max || likely < min )
+            {
+                throw new ArgumentException(Resources.InvalidDistributionParameters);
+            }
+
+            // specified to make the formulas match the literature;
+            // traditionally set to 4 so that the range between min and max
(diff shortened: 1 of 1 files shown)


cdrnet:
Great, thanks!

For API consistency it would be nice if it would optionally accept a random source. Could you add `System.Random randomSource=null` as optional last argument and just pass the value down to the constructor?

cdrnet:
Thanks

david-in-perth:
@jbowtie : Great idea. We could use that at my work, too.

It doesn't seem right as-is, though. The constructor accepts min, max, and mode, but returns a two-parameter Beta distribution, which has a range of [0,1]. So, for example, the distribution object returned by PERT(50, 65, 60) still has to be transformed to give values between 50 and 65.

A *four-parameter* Beta distribution, which has translation and scaling built-in, would be useful here.

Example references (from a quick Google search):

- https://www.me.utexas.edu/~jensen/ORMM/omie/computation/unit/project/beta.html
- http://math.stackexchange.com/questions/179537/four-parameter-beta-distribution-and-wikipedia
- http://www.itl.nist.gov/div898/handbook/eda/section3/eda366h.htm

david-in-perth:
Also, lambda could be added as a fourth (optional) distribution parameter, with a default of 4.

jbowtie:
I haven't seen cases where lambda is anything other than 4 but making it an optional parameter is a possibility in a future pull request.

Creating a four-parameter Beta distribution seems like a more ambitious undertaking suited to its own issue.

david-in-perth:
> I haven't seen cases where lambda is anything other than 4 but making it an optional parameter is a possibility in a future pull request.

Yeah, I have no information to suggest that lambda values other than 4 are commonly used, so that's probably just a 'nice-to-have' / not important.

> Creating a four-parameter Beta distribution seems like a more ambitious undertaking suited to its own issue.

It might be less ambitious than you think, but I agree that it deserves its own issue, which I've created (#322). I also created an issue for the problem with the PERT method, as mentioned above (#321).

PR #333, opened 2015-07-05
Added RootFinding.Secant and unit tests (based on NewtonRaphson)

grovesNL:
The unit tests had to be modified slightly, because the Secant method supports two initial guesses while still maintaining lowerBound and upperBound. It's possible to remove the need for initial guesses entirely and rely on the bounds, but the stability of the Secant method could cause it to diverge outside of the bounds immediately.

I could add additional tests to for Secant failure specifically if desired. Due to the nature of Secant, it will fail on repeated roots, or when the interval is split perfectly around a root, etc.

### DIFF
--- a/src/Numerics/Numerics.csproj
+++ b/src/Numerics/Numerics.csproj
@@ -210,6 +210,7 @@
     <Compile Include="RootFinding\Cubic.cs" />
     <Compile Include="RootFinding\NewtonRaphson.cs" />
     <Compile Include="RootFinding\RobustNewtonRaphson.cs" />
+    <Compile Include="RootFinding\Secant.cs" />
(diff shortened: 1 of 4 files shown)


cdrnet:
Great, thanks!

grovesNL:
@cdrnet: It seems that my commit had a typo in the author email address ("grgoves" instead of "groves"). Can this be easily changed? It's not a big concern regardless.

cdrnet:
@grovesNL Technically it could be fixed but I prefer not to rewrite it at this point. But I can add an entry to the mailrc file so at least the git log commands would return the correct email address.

grovesNL:
@cdrnet: Perfect, thanks!

PR #336, opened 2015-07-13
Added Functional Vector and Matrix Integration

MadcowD:
Created extensions for Vector and Matrix integration in Numerics.Integrate. Should I add these directly to the individual integration methods and thereafter create an extension within Numerics.Integrate?

### DIFF
--- a/src/Numerics/Integrate.cs
+++ b/src/Numerics/Integrate.cs
@@ -30,6 +30,7 @@
 
 using System;
 using MathNet.Numerics.Integration;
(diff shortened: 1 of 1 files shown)


cdrnet:
This looks interesting - and is yet another case where we'd have a need for more generic non-LA matrix/vector types (the other one is integer matrices e.g. for indexing)

MadcowD:
Would you mind if I explored this domain?

-----Original Message-----

Issue #327, opened 2015-06-12
MKL: lapack info variable is not handled correct


tibel:
In multiple function/templates in `lapack.cpp` the value of `info` is not checked before continuing with other operations.

tibel:
Also this needs to be checked in managed code, as only a value `zero´ means success (see https://software.intel.com/de-de/node/520868).

cuda:
I started going through the `info` return values. What about cases such as `getrf`:
```
If info = 0, the execution is successful.
If info = -i, the i-th parameter had an illegal value.
If info = i, uii is 0. The factorization has been completed, but U is exactly singular. 
Division by 0 will occur if you use the factor U for solving a system of linear equations.
```
In this case, I don't think we should throw an exception if `info > 0`, no?

tibel:
You are right, in this case it depends on what operation comes next (we can handle this later).

cuda:
sorry for the delay, not much spare time lately...I pushed the error checking for LU factor and inverse as an example..
https://github.com/cuda/mathnet-numerics/tree/memory

tibel:
Thanks.

Issue #326, opened 2015-06-12
MKL: Reason of allocating work array in managed code


tibel:
Why the work arrays for LAPACK (lu, qr, svd) are created in C# (managed code)?

I think they are needed only for the calculation (temporarily). When allocating them in native code we could benefit from the MKL memory pools and reduce stress of CLR memory allocator and GC.

tibel:
The `work` array is an implementation detail of the provider and should not be part of the public interface. 
``` C#
ILinearAlgebraProvider<T>
{
  void LUInverse(T[] a, int order, T[] work);
  void LUInverseFactored(T[] a, int order, int[] ipiv, T[] work);
  void QRFactor(T[] a, int rowsA, int columnsA, T[] q, T[] tau, T[] work);
  void ThinQRFactor(T[] a, int rowsA, int columnsA, T[] r, T[] tau, T[] work);
  void QRSolve(T[] a, int rows, int columns, T[] b, int columnsB, T[] x, T[] work, QRMethod method = QRMethod.Full);
  void QRSolveFactored(T[] q, T[] r, int rowsA, int columnsA, T[] tau, T[] b, int columnsB, T[] x, T[] work, QRMethod method = QRMethod.Full);
  void SingularValueDecomposition(bool computeVectors, T[] a, int rowsA, int columnsA, T[] s, T[] u, T[] vt, T[] work);
}
```

When using CUDA the work array has to be allocated in graphics memory.

cuda:
The original idea was to support creating a single work array in .NET, and use it over and over, instead of allocating one each time.  Yes, GPU providers couldn't use it, but they didn't need to support it (the same way the managed provider wouldn't in many cases). But I don't think we ever provided a way from the managed code to support reusing the a work array. So, I'm all for removing it from the public interface and requiring the providers to handle it themselves. I'll work on the MKL provider next week.

tibel:
Thanks for fixing this.

Issue #325, opened 2015-06-11
Usage of new/delete in native code


tibel:
I think we should not use `new` and `delete` operator in the native providers.
When `new` cannot allocate the memory a `std::bad_alloc` is thrown.

.NET _DllImport_ uses _C export_ and throwing C++ errors over a C ABI is not nice.
E.g. for MKL `mkl_malloc` and `mkl_free` could be used instead (https://software.intel.com/en-us/node/522107).

cuda:
great suggestion!

tibel:
- [ ] Common (`Clone` in `lapack_common.h`)
- [ ] ATLAS
- [ ] CUDA
- [ ] MKL
- [ ] OpenBLAS

tibel:
@cuda I haven't done C/C++ development for a long time. Adding more exit (return) points to the methods will make them less readable and harder to maintain. Do you have an idea how to solve this (e.g. using goto statements)?

cuda:
I have an idea, I'll tinker with it in a couple days.

tibel:
My version (with lack of better C/C++ knowledge) for MKL: 
https://gist.github.com/tibel/9f1de1447d4bb01670c9

cuda:
that would work, but I wonder if we could use unique_ptr with a custom deleter and exceptions (but handle them inside the native library)
maybe something like this (compiles, but not really tested)
https://gist.github.com/cuda/3b164db015cf00c4580c

tibel:
I think that the MKL functions do not throw any exception.
Do you think it is needed to introduce exceptions for MathNet memory allocations only?

cuda:
>I think that the MKL functions do not throw any exception.
they don't, I threw one if the allocation failed.

>Do you think it is needed to introduce exceptions for MathNet memory allocations only?
In my opinion, yes. I think the code is cleaner with exceptions.

kjbartel:
This is also related to the #316 changes which are currently using an approach similar to @tibel did in his gist.

cuda:
Thanks, I haven't looked at CUDA provider. Whatever approach you guys thinks is best.

kjbartel:
> Whatever approach you guys thinks is best.

I was thinking the same thing. I just wanted to point out that the CUDA provider recently had similar changes for handling memory allocation errors. So whichever approach is used should be consistent for all the native providers.

[...]

PR #338, opened 2015-07-25
Cleaning up native memory usage and other minor issues

cuda:
This addresses issues #325, #326 and #327.


cdrnet:
Great, thanks again Marcus! Sorry for the delay, just back from vacation...

Quick question: was the change to drop the wildcard file inclusion approach in UnitTests-MKL.csproj intentional?

cuda:
> was the change to drop the wildcard file inclusion approach in UnitTests-MKL.csproj intentional?

No, that was a mistake. Can easily revert it?

cdrnet:
Ok. Yes, I can revert just that file easily after pulling.

cuda:
OK, thanks.

PR #340, opened 2015-08-30
Nelder-Mead simplex algorithm (unified optimization)

eriove:
I needed the Nelder-Mead simplex algorithm and found an MIT licensed version [here](https://code.google.com/p/nelder-mead-simplex/). I've adapted this to fit into the unified-optimization branch. Maybe not the best idea to add something to a branch that is "work-in-progress" but since it was a quite small addition I thought it was worth the effort anyway :-).

The Nelder-Mead simplex algorithm is probably mainly known as [fminsearch()](http://se.mathworks.com/help/matlab/math/optimizing-nonlinear-functions.html#bsgpq6p-11) in Matlab and is quite useful when you need to do a minimization and don't have access to the gradient/or if it's expensive to calculate.


cdrnet:
Great, thanks a lot!

The remaining test failure reported by AppVeyor is unrelated to these changes.

PR #342, opened 2015-09-30
Minor comment corrections

BenHewins:
I noticed (while adding a truncated normal distribution) that a bunch of comments where slightly wrong. I don't think anyone would have been confused by them but fixed them anyway.

Comments on some properties on most Distributions incorrectly stated
that they were setters as well.
One comment on an array statistic stated it was returned the smallest
instead of largest member.


cdrnet:
Thanks a lot for these corrections!

The mentioned properties used to have a setter in v2, but it seems the comments were not updated correctly when the setters were removed.

PR #345, opened 2015-10-18
Update mkl native provider for OSX

lionpeloux:
Hello,

I've updated the mkl_build script to build the MKL native provider on OSX thanks to directions from @cuda and @cdrnet. I've succeed in building the native provider on my mac and run the example given in the doc (matrix multiplication).

I've also enriched the OSX section of the doc file (MKL.md) so people not familiar with the  build process (like me) can do it themselves more easily.

Lionel

PS : I'm running OSX 10.10.5 with intel Parallel Studio XE 2016 on a macbook pro end 2011 - core i7 2.8 GHz.


cdrnet:
Thanks a lot for improving the build script, and the docs as well!

Issue #323, opened 2015-06-10
Histogram: Adding data which includes the lower bound


vkb:
Hello,

i run into a problem with data that includes the exact value of the lower bound of a histogram. 

Testcase:

        [Test]
        public void AddDataIncludeBoundaries()
        {
            Assert.DoesNotThrow(() => new Histogram(new[] { 0.0 }, 2, 0.0, 10.0));
        }

In that case an Exception is thrown. If i try to add a lower value, like '-1.0', the bucket of the lowest order is extended. Therefore i assume this is not a desired behavior.

One solution would be to modify the Contains method of the Bucket class:

        public int Contains(double x)
        {
            if (LowerBound <= x)
            {
                if (UpperBound >= x)
                {
                    return 0;
                }

                return 1;
            }

            return -1;
        }

I replaced 'LowerBound < x' with 'LowerBound <= x'. Otherwise the value 'x' is included if it is smaller or **equal** the upper bound, but ***not*** if it is greater or **equal** the lower bound. This might be desired, as the next bucket shell include it's upper, but not it's lower bound because that one is included in the previous one. This problem is unique to the first bucket.

Alternatively the 'AddData()' method of the 'Histogram' class needs to be modified:

        public void AddData(double d)
        {
            // Sort if needed.
            LazySort();

            if (d <= LowerBound)
            {
                // Make the lower bound just slightly smaller than the datapoint so it is contained in this bucket.
                _buckets[0].LowerBound = d.Decrement();
                _buckets[0].Count++;
            }
            else if (d > UpperBound)
            {
                _buckets[BucketCount - 1].UpperBound = d;
                _buckets[BucketCount - 1].Count++;
            }
            else
            {
                _buckets[GetBucketIndexOf(d)].Count++;
            }
        }

I replaced 'd < LowerBound' with 'd <= LowerBound'. This would extend the lower bound to include the value.

I'm eager to hear your opinion. Thank you for your work!

regards 
Volker

cdrnet:
Fixed in #353 - thanks again!

PR #353, opened 2015-11-19
Histogram: Adding data equal to the lower bound

vkb:
As described in issue https://github.com/mathnet/mathnet-numerics/issues/323, adding values that are equal to the lower bound of a histogram result in an exception. Values slightly lower than that bound, will decrease the lower bound automatically as required.

Therefore, i propose to initiate this behavior on all values *equal* or lower that the boundary.


cdrnet:
Thanks!

PR #366, opened 2016-01-23
More accurate algorithms for Complex

MaLiN2223:
I changed magnitude because of overflow for big numbers, It's now based on simple mathematical observation that Sqrt(a^2+b^2) = |a| * Sqrt(1+b/a*b/a). 
Problem with squaring might occur for example when we had a or b bigger than (2 *10^19) because of (2 *10^19)^2 == 4 *10^38 > 3.4 *10^38 == float.MaxValue.  

Second change is in division. Now there will be no Nan or overflow when dividing two numbers with real/imaginary values near float.epsilon. It's based on Smith's algorithm but is enhanced because Smith's sometimes returned bad imaginary values. 

For more examples see unit test in my commit.

### DIFF
--- a/src/Numerics/Complex32.cs
+++ b/src/Numerics/Complex32.cs
@@ -180,7 +180,7 @@ namespace MathNet.Numerics
         /// <summary>
         /// Gets the magnitude (or absolute value) of a complex number.
         /// </summary>
-        /// <remarks>Assuming that magnitude of (inf,a) and (a,inf) is NaN and magnitude of (inf,inf) is inf</remarks>
+        /// <remarks>Assuming that magnitude of (inf,a) and (a,inf) and (inf,inf) is inf and (NaN,a), (a,NaN) and (NaN,NaN) is NaN</remarks>
         /// <returns>The magnitude of the current instance.</returns>
         public float Magnitude
         {
--- a/src/Numerics/Complex32.cs
+++ b/src/Numerics/Complex32.cs
@@ -184,7 +184,26 @@ namespace MathNet.Numerics
         public float Magnitude
         {
             [TargetedPatchingOptOut("Performance critical to inline this type of method across NGen image boundaries")]
-            get { return (float)Math.Sqrt((_real * _real) + (_imag * _imag)); }
+            get
+            {
+                float a = Math.Abs(_real);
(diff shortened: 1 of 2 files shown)


cdrnet:
Great, that looks like good improvement. Thanks!

src/Numerics/Complex32.cs
```diff
@@ -620,6 +641,7 @@ public Complex32 SquareRoot()
         }
 
         /// <summary>Division operator. Divides a complex number by another.</summary>
+        /// <remarks>Enchanted Smith's algorithm for dividing two complex numbers </remarks>
```
cdrnet: Is this intended to read "Enchanted" or should it rather be "Enhanced"? ;)

src/Numerics/Complex32.cs
```diff
@@ -184,7 +184,28 @@ public float Phase
         public float Magnitude
         {
             [TargetedPatchingOptOut("Performance critical to inline this type of method across NGen image boundaries")]
-            get { return (float)Math.Sqrt((_real * _real) + (_imag * _imag)); }
+            get
+            {
+                float a = Math.Abs(_real);
+                float b = Math.Abs(_imag);
+                if (float.IsPositiveInfinity(a) && float.IsPositiveInfinity(b))
```
cdrnet: Wouldn't this even hold with an `||` instead of a `&&`, i.e. if just one of them is infinite?

src/Numerics/Complex32.cs
MaLiN2223: Of course it should be "Enhanced" ;)

src/Numerics/Complex32.cs
MaLiN2223: I think it strongly depends on the definition. We have to ask someone who uses infinite complex numbers. Shall I change it for || ? And also, yes this algorithm should hold double interpretation.

MaLiN2223:
I recently found that there is float/complex and complex/float overflowing problem so please do not merge my branch with master just yet. Let me correct this first.

cdrnet:
Thanks for the updates!

cdrnet:
Great, thanks again!

PR #367, opened 2016-02-05
Added Quaternion and QuaternionTests to the FSharp project

pdoh00:
Here is an FSharp implementation of a Quaternion.  Please let me know if I am missing anything according to contribution guidelines.  I have tried to include references for where I found my math.

### DIFF
--- a/src/FSharp/FSharp-Net35.fsproj
+++ b/src/FSharp/FSharp-Net35.fsproj
@@ -75,6 +75,7 @@
     <Compile Include="FindRoots.fs" />
     <Compile Include="RandomVariable.fs" />
     <None Include="MathNet.Numerics.fsx" />
+    <Compile Include="Quaternion.fs" />
   </ItemGroup>
   <ItemGroup>
     <Reference Include="FSharp.Core, Version=2.3.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
--- a/src/FSharp/FSharp.fsproj
+++ b/src/FSharp/FSharp.fsproj
@@ -76,6 +76,7 @@
     <Compile Include="RandomVariable.fs" />
     <None Include="MathNet.Numerics.fsx" />
     <None Include="MathNet.Numerics.IfSharp.fsx" />
+    <Compile Include="Quaternion.fs" />
   </ItemGroup>
   <ItemGroup>
     <Reference Include="mscorlib" />
--- a/src/FSharp/Quaternion.fs
+++ b/src/FSharp/Quaternion.fs
@@ -1,0 +1,90 @@
+﻿module MathNet.Numerics.Quaternion
+
+//Reference:
+//http://www.astro.rug.nl/software/kapteyn/_downloads/attitude.pdf
+//http://www.mathworks.com/help/aeroblks/quaternionmultiplication.html
+//http://www.mathworks.com/help/aeroblks/quaterniondivision.html
+//https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation#The_conjugation_operation
+
+type Quaternion = 
+    {
+        w:float
+        x:float
+        y:float
+        z:float
+    } with
+    static member (+) (r: Quaternion, q: Quaternion) =
+        {w=r.w+q.w;x=r.x+q.x;y=r.y+q.y;z=r.z+q.z}
+    static member (-) (r: Quaternion, q: Quaternion) =
+        {w=r.w-q.w;x=r.x-q.x;y=r.y-q.y;z=r.z-q.z}
+    static member (*) (r: Quaternion, q: Quaternion) =
+        let w = r.w*q.w - r.x*q.x - r.y*q.y - r.z*q.z
+        let x = r.w*q.x + r.x*q.w - r.y*q.z + r.z*q.y
+        let y = r.w*q.y + r.x*q.z + r.y*q.w - r.z*q.x
+        let z = r.w*q.z - r.x*q.y + r.y*q.x + r.z*q.w
+        {w=w;x=x;y=y;z=z}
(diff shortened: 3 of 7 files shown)


kjbartel:
Wouldn't it be better to implement this in C# so that all of MathNet.Numerics users could use it rather than just F# users?

pdoh00:
FSharp is a full fledged member of the .NET run time, so this should be usable from C#.

kjbartel:
Yes. But that is not what I'm saying. The `MathNet.Numerics` assembly is the "core package" while the `MathNet.Numerics.FSharp` assembly contains "optional extensions for a better F# experience". People shouldn't have to include the `MathNet.Numerics.FSharp` assembly to use this from C#.

pdoh00:
I can sympathize with your point of view, since I only recently started writing F# and have been writing C# for most of my career.  I don't think it's valid to not accept an F# contribution because the library was started when F# wasn't a viable option.

I would like to think :pray: that if F# was mature/existed when Math.net started, "core" would be in F#.  I believe that F# is a superior language for succinctly expressing mathematical based code. This is objectively true, for this class, by comparing to https://github.com/dotnet/corefx/blob/master/src/System.Numerics.Vectors/src/System/Numerics/Quaternion.cs. While I didn't implement everything that Numerics does, it's plain to see the difference in the methods that I did implement.

kjbartel:
> I would like to think :pray: that if F# was mature/existed when Math.net started, "core" would be in F#

That may be true, but currently `MathNet.Numerics` is written in C# so something which should be available to all .Net languages I think should be in the core assembly and hence written in C#.

It may be the case that for future versions of MathNet.Numerics the code base is moved to F#. It is also possible to sort of mix languages by using ILMerge to merge multiple assemblies together into a single assembly. That could allow sections to be migrated to F# gradually and new functionality to be written in F#. But it may not be feasible as merged assemblies have to be independent of each other (or at least the C# assembly would not be able to call the (new) F# assembly, meaning only non-core sections could be moved to F#).

If the project were to go down that path though I think it would need to be a clear that that is the desired development direction.

cdrnet:
Thanks!

We already have some functionality available only in the FSharp package that may also be useful for non-F# .net developers, mostly the `BigRational` type. So it is acceptable in my point of view to do the same with `Quaternion`.

But I do agree that the current separation between the two packages is unfortunate. A separation into two NuGet packages just because it uses a different language internally doesn't make any sense to users. One problem with a single package approach could be the dependency on `FSharp.Core` though, which could cause some problems.

kjbartel:
> One problem with a single package approach could be the dependency on FSharp.Core though, which could cause some problems.

Would that really be a problem? If `MathNet.Numerics` was completely written in F# then it would have the same dependency. Or if people use features only available in the `MathNet.Numerics.FSharp` assembly. I can't think of any issues this would cause especially seeing as `FSharp.Core` is available through a NuGet package which could be updated independently of `MathNet.Numerics`.

cdrnet:
FSharp.Core used to cause quite a few problems mostly due to it sitting in the GAC and having special treatment in the tooling. Luckily this is about to be getting better: no more GAC and it is becoming standard to use the proper FSharp.Core NuGet package.

For example, while LINQPad supports F#, it still treats FSharp.Core specially and cannot handle NuGet packages that depend on the FSharp.Core NuGet package (like MathNet.Numerics.FSharp does). This means that while it is possible to use our F# package from C# in LINQPad, and the MathNet.Numerics package in F#, it is currently not possible to use the F# package from F#.

cdrnet:
@pdoh00 do you require the `norm` function to be defined as is, or can I rename it to `normSquared` and add a new `norm` function that returns the square root of that? Most sources seem to agree on the square root. MathWorks/MATLAB disagrees, but to my understanding their definition actually violates the subadditivity property of norms and thus does not actually form a proper norm.

cdrnet:
Also, does it make sense in practice that `rotate` returns a quaternion?

pdoh00:
I actually changed norm due to the way MatLab has it because they had test data. I am not against changing it. Originally I implemented it the same as the referenced paper. 

It might be more appropriate to return a Vector3 from rotate. The reason I did not do that originally was that I was test a pure solution without pull Math Numerics in.  Now that Quaternion is part of numerics it might make more sense to return a Vector3...maybe even take a Vector3 instead of x,y,z.

:+1:

pdoh00:
@cdrnet I am just posting some code here with some ideas on the Quaternion methods for create and rotate. I am doing some like this internally at work.

```
let create (angle:float<Radian>) (v : Vector3<'u>) =
    //axis needs to be unit vector
    let v' = Vector3.normalize v
    let halfAngle = angle * 0.5
    //divide by 1.0<radian> to get back to dimensionless value for sin and cos
    let s = halfAngle / 1.0<Radian> |> sin
    let c = halfAngle / 1.0<Radian> |> cos
    {
      W=c;
      X= v'.X / (LanguagePrimitives.FloatWithMeasure 1.0) * s;
      Y= v'.Y / (LanguagePrimitives.FloatWithMeasure 1.0) *s;
      Z= v'.Z / (LanguagePrimitives.FloatWithMeasure 1.0) *s
    }

 let rotate (q:Quaternion) (v : Vector3<'u>) =
    let qn = normalize q
    let p = {W= 0.0; X= float v.X; Y= float v.Y; Z= float v.Z}
    let rotated = qn * p * inverse qn
    {
      Vector3.X=rotated.X* (LanguagePrimitives.FloatWithMeasure 1.0);
      Y=rotated.Y* (LanguagePrimitives.FloatWithMeasure 1.0);
      Z=rotated.Z* (LanguagePrimitives.FloatWithMeasure 1.0)
    }
```

Obviously for ```Numerics``` it would use DenseVector or the like. Since this code only works for Vectors of size 3 I am not totally sure how to enforce that with DenseVector.  I hope this helps.

PR #371, opened 2016-02-09
Gauss-Legendre Rule Feature

larzw:
- Support for obtaining the Gauss-Legendre abscissas/weights
- Integration in 1D and 2D.
- Added Unit tests


cdrnet:
That looks great, thanks!

Should we also expose this in the `MathNet.Numerics.Integrate` facade class (which focuses on simple use cases rather than algorithms)? Maybe the 2D integration?

src/UnitTests/IntegrationTests/IntegrationTest.cs
```diff
@@ -47,6 +47,17 @@ private static double TargetFunctionA(double x)
         }
 
         /// <summary>
+        /// Test Function: f(x,y) = exp(-x/5) (2 + sin(x * y))
+        /// </summary>
+        /// <param name="x">First input value.</param>
+        /// <param name="y">Second input value.</param>
+        /// <returns>Function result.</returns>
+        private static double TargeFunctionB(double x, double y)
```
cdrnet: Small typo, missing a `t`.

src/UnitTests/UnitTests.csproj
```diff
@@ -65,6 +65,10 @@
     <LangVersion>5</LangVersion>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="nunit.framework, Version=3.0.5813.39035, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\NUnit\lib\nunit.framework.dll</HintPath>
+    </Reference>
     <Reference Include="System" />
```
cdrnet: I don't think this reference should be changed. Maybe my recent changes to use paket dependency groups (NuGet) have confused some tooling?

src/Numerics/Integration/GaussLegendreRule.cs
```diff
@@ -0,0 +1,263 @@
...
+        /// Approximates a definite integral using an Nth order Gauss-Legendre rule.
+        /// </summary>
+        /// <param name="f">The analytic smooth function to integrate.</param>
+        /// <param name="invervalBegin">Where the interval starts, exclusive and finite.</param>
+        /// <param name="invervalEnd">Where the interval ends, exclusive and finite.</param>
+        /// <param name="order">Defines an Nth order Gauss-Legendre rule. The order also defines the number of abscissas and weights for the rule. Precomputed Gauss-Legendre abscissas/weights for orders 2,. . ., 20, 32, 64, 96, 100, 128, 256, 512, 1024 are used, otherwise they're calulcated on the fly.</param>
+        /// <returns>Approximation of the finite integral in the given interval.</returns>
+        public static double Integrate(Func<double, double> f, double invervalBegin, double invervalEnd, int order)
+        {
+            nonNegativeGaussLegendrePoint = GetGaussPoint(order);
```
cdrnet: The use of the static field here looks to me like it could break when used in parallel from two threads. I assume the idea is to reuse it, as the order will likely remain the same between evaluations?

cdrnet: Should maybe the factory do the caching instead of the rule itself? This would make the rule stateless.

larzw:
Hi! thanks for the quick response.

I fixed the issue with modifying src/UnitTests/UnitTests.csproj. It's now the same as master. I'm still working on the others. I was also thinking of adding examples to src/Examples/Integration.cs, if you think that's a good idea?

larzw:
I believe I've address all your comments.

cdrnet:
Great, thanks a lot for the update!

PR #374, opened 2016-02-10
Moved initialization of LinearAlgebraProvider to reduce exceptions

kjbartel:
Moved initialization of default LinearAlgebraProvider out of Control.ConfigureAuto to prevent unnecessary initialization and exceptions. Initialization code moved to new method, InitializeDefaultLinearAlgebraProvider(), which is called from the LinearAlgebraProvider getter only if the current _linearAlgebraProvider is null. This prevents trying to initialize multiple providers when calling any of the 'UseXX' methods.


cdrnet:
Thanks!

PR #376, opened 2016-02-14
Added quaternion in C#

MaLiN2223:
[THIS PULL IS NOT FULLY READY YET]

I noticed that there is implementation of quaternions in F# : [here](https://github.com/mathnet/mathnet-numerics/pull/367)

Kjbartel said that we need C# implementation of it thus I started to implement.

Just give me one or two days to make it fully work.

Things to do :

Actual Code : 
- [ ] Rotations
- [ ] More robust division (possible overflow)
- [ ] Create more robust norm (possible overflow)
- [ ] Create quaternion from angle
- [ ] Create quaternion from matrix

--
Testing
- [ ] DotProduct
- [ ] Inversion
- [ ] Equality, operator ==
- [ ] ToString()
- [ ] Norm
- [ ] Conjugation
- [ ] operator* for floats 
- [ ] operator/

If you have any more ideas what to add just post it - I will handle it.
If you want to contribute :  [here is my git repo](https://github.com/MaLiN2223/mathnet-numerics/tree/malins)

### DIFF
--- a/src/Numerics/Numerics.csproj
+++ b/src/Numerics/Numerics.csproj
@@ -214,6 +214,7 @@
     <Compile Include="Exceptions.cs" />
     <Compile Include="Providers\LinearAlgebra\OpenBlas\OpenBlasProviderCapabilities.cs" />
     <Compile Include="Providers\NativeProviderLoader.cs" />
+    <Compile Include="Quaternion32.cs" />
     <Compile Include="Random\SystemRandomSource.cs" />
     <Compile Include="Random\RandomSeed.cs" />
     <Compile Include="RootFinding\Broyden.cs" />
--- a/src/Numerics/Quaternion32.cs
+++ b/src/Numerics/Quaternion32.cs
@@ -1,0 +1,145 @@
+﻿// <copyright file="Quaternion.cs" company="Math.NET">
+// Math.NET Numerics, part of the Math.NET Project
+// http://numerics.mathdotnet.com
+// http://github.com/mathnet/mathnet-numerics
+// http://mathnetnumerics.codeplex.com
+//
+// Copyright (c) 2009-2010 Math.NET
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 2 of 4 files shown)


kjbartel:
> Kjbartel said that we need C# implementation of it thus I started to implement.

Just because said something doesn't make it true. :) I may very well have been mistaken in my understanding of the purpose of the `MathNet.Numerics.FSharp` assembly.

If it is supposed to be _**only**_ for use in F# then _**I think**_ that everything should be included in the normal `MathNet.Numerics` assembly. But, if the `MathNet.Numerics.FSharp` assembly is simply supposed to be additional functionality and utilities, not part of the core `Numerics` functionality, for general use from all .Net languages (which just happen to be implemented in F#) then there may be no need for this to be implemented in the core `MathNet.Numerics` assembly.

This comment from @cdrnet makes me think it is the latter:

> We already have some functionality available only in the FSharp package that may also be useful for non-F# .net developers, mostly the BigRational type. So it is acceptable in my point of view to do the same with Quaternion.

> But I do agree that the current separation between the two packages is unfortunate. A separation into two NuGet packages just because it uses a different language internally doesn't make any sense to users. One problem with a single package approach could be the dependency on FSharp.Core though, which could cause some problems.

So rather than implementing this in C# I think it would be better to look at removing the confusion of the purpose of the `MathNet.Numerics.FSharp` assembly / package. The easiest would perhaps be simple renaming it to something like `MathNet.Numerics.Extras`. But I think there are things in the assembly which are specifically targeted at F#. So it may be better to even create an additional assembly for the non-F# utilities.

Indeed the whole `Numerics` project could be broken up into smaller assemblies and packages focused on specific areas / fields. That may also allow for some areas to be re-implemented in F# without needing to re-implement the whole library. Whether it's worth the effort or not though isn't for me to say. The only real benefits are being able to have smaller assemblies (which may be desirable for mobile and embedded use) and for separate versioning and release of assemblies, allowing functionality to be released faster.

The opposite direction would be looking at ways to create a single NuGet package and / or assembly. The two assemblies could easily be packaged together or could be merged into a single assembly with ILmerge. But, as @cdrnet said there may be issues with an `FSharp.Core` dependency neither of these options may be desirable, at least at the moment.

Bottom line: Not sure if this is needed or not.

MaLiN2223:
I just liked your idea to do that :( And since I like to create C# programs I did it with pleasure.
Okay, if you don't know if it is necessary I'll wais for what @cdrnet say about this idea. Should I start a discussion on codeplex?

cdrnet:
@MaLiN2223 would you also be interested in contributing this to [Math.NET Spatial](https://github.com/mathnet/mathnet-spatial) instead? Spatial is also part of Math.NET and sits on top of Numerics. It provides types and functionality related to spatial problems (geometry, ..) and already has a Quaternion type, although one that is implemented quite badly (it precomputes certain properties which doesn't make much sense, its norm implementation is wrong, its API quite weird, etc.), and already knows concepts like Euler-angles. It would be great if we'd get some help to bring Spatial finally into a releasable state.

MaLiN2223:
Of course I would! 
Thank you. I am closing this pull request right now and start to decoding Spatial :)

PR #385, opened 2016-03-30
Fix indirect dependencies on native dlls

mjmckp:
Suppose Lib1 is a C# project which uses the NuGet package containing the native dll, and Lib2 depends on Lib1, and finally Lib3 depends on Lib2.  With the current setup, the native dlls will be copied into the output directory of both Lib1 and Lib2, but not Lib3.  Using 'None' instead of 'Content' fixes this.

I don't have a reference for this behaviour, but I have replicated it with three dummy projects to be sure.  I originally got the hint from Benoit Blanchon's comment at http://stackoverflow.com/questions/19478775/


cdrnet:
Thanks, this looks interesting! I'll need to do some testing but will likely merge.

mjmckp:
@cdrnet This change requires a release of the `MathNet.Numerics.MKL.Win-x64` NuGet package, would it be possible arrange this, please?

PR #389, opened 2016-04-11
improve memory usage of parsing.

OpportunityLiu:
before

after

test file is a 3700*3700 matrix, 385MB.


cdrnet:
That looks much better indeed, thanks!

cdrnet:
I've scheduled a release of the Data NuGet package later today.

cdrnet:
Released in data extensions v3.2.0.

PR #399, opened 2016-05-26
Update GaussLegendrePointFactory.cs

skosukhin:
Typo fix. Now it is possible to create an instance of GaussLegendreRule with 256th order.


cdrnet:
Great, thanks!

PR #403, opened 2016-06-04
Fixed accuracy handling in RootFinding.Bisection

konstantint:
At the moment Bisection.FindRoot would iterate until it converges on the maximally precise answer, basically ignoring the "accuracy" attribute. I am using this routine to find a root of a function which is computationally complex, hence every extra invocation matters in my case.

The reason for the incorrect handling of the accuracy lies in the rather strange code at line 111 of the current Bisection.cs:

```
if (Math.Abs(fmax - fmin) < 0.5*accuracy && upperBound.AlmostEqualRelative(lowerBound))
```

As you see this requires the algorithm to run all the way until upperBound equals lowerBound, not until the requested accuracy margin is achieved.

There is another problem: the function makes two accuracy tests at the very beginning. The tests are of the following type:

```
if (Math.Abs(fmin) < accuracy)
```

This implies that the "accuracy" parameter is interpreted in terms of the function value rather than the accuracy of the located root. I find this to be contrary to general intuition. For example, the following test would fail at the moment:

```
Assert.AreEqual(0.5, Bisection.FindRoot(x => 1e-5*(x - 0.5), 0.0, 1.0, 1e-5), 1e-5);
```

Obviously, the fix would be to replace the current termination criteria with the following:

```
if (upperBound - lowerBound <= accuracy) return true;
```

Of course, taking into account the "accuracy of y" might also be important - otherwise the function would return as root any point where the function jumps over zero without actually touching it, hence a more reasonable change would be:

```
if (upperBound - lowerBound <= accuracy && Math.Abs(froot) <= accuracy) return true;
```

(Ideally, there might be a separate parameter y-accuracy, but adding a new parameter looks like an overkill at the moment).

Making the necessary changes, however, breaks many tests. Firstly, some tests require infinite precision:

```
 Assert.AreEqual(0, f1(Bisection.FindRoot(f1, 0, 5, 1e-14)));
```

Secondly, many tests require more precision than what is specified in the FindRoot invocation.

In thi PR I'm suggesting to change:
- Termination handling to correctly respect accuracy as described above, avoiding unnecessary iterations and respecting accuracy on the x axis as well as y.
- Fix tests by adding appropriate tolerance in the comparisons.
- Change the default accuracy parameter value to 1e-14 - this way most of the original tests still pass and the change is hopefully not too breaking for users which rely on those default values.

Note that it still is a breaking change for some tests - the fix requires manual specification of a smaller accuracy threshold than 1e-14 (otherwise the new algorithm does not find a solution at all).

### DIFF
--- a/src/FSharpUnitTests/FindRootsTests.fs
+++ b/src/FSharpUnitTests/FindRootsTests.fs
@@ -21,8 +21,14 @@ module FindRootsTests =
 
     [<Test>]
     let ``Bisection should find both roots of (x - 3) * (x - 4)``() =
-        f |> FindRoots.bisection 100 1e-14 -5.0 3.5 |> should (equalWithin 1e-14) 3.0
-        f |> FindRoots.bisection 100 1e-14 3.2 5.0 |> should (equalWithin 1e-14) 4.0
+        f |> FindRoots.bisection 100 1e-14 -5.0 3.5 |> (fun x ->
+            match x with
+            | Some x -> should (equalWithin 1e-14) 3.0
+            | None ->   failwith "The element is not equal.") |> ignore
+        f |> FindRoots.bisection 100 1e-14 3.2 5.0 |> (fun x ->
+            match x with
+            | Some x -> should (equalWithin 1e-14) 4.0
+            | None ->   failwith "The element is not equal.") |> ignore
 
     [<Test>]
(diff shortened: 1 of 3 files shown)


cdrnet:
Thanks a lot for this analysis!

cdrnet:
Finally merged - thanks again!

PR #404, opened 2016-06-08
Merged BFGS solvers and line search algorithms

eriove:
As mentioned in #283 I tried to merge as much as possible of the BFGS algorithms. Also merged part of the WolfeLineSearch algorithms.

Found a bug in a test of the Nelder Mead Simplex algorithm, had to remove the test since it was the  only reason for the test to pass.

I'm a little worried that almost all optimization algorithms only are tested with the Rosenbrock function but I'm usually not working with minimization problems that has easily available gradients so I'm not sure what to test them with.


src/Numerics/Optimization/BfgsBMinimizer.cs
```diff
@@ -55,9 +92,9 @@ public MinimizationResult FindMinimum(IObjectiveFunction objective, Vector<doubl
...
             var gradientProjectionResult = QuadraticGradientProjectionSearch.Search(objective.Point, objective.Gradient, pseudoHessian, lowerBound, upperBound);
-            var cauchyPoint = gradientProjectionResult.Item1;
-            var fixedCount = gradientProjectionResult.Item2;
-            var isFixed = gradientProjectionResult.Item3;
+            var cauchyPoint = gradientProjectionResult.CauchyPoint;
+            var fixedCount = gradientProjectionResult.FixedCount;
+            var isFixed = gradientProjectionResult.IsFixed;
             var freeCount = lowerBound.Count - fixedCount;
 
             if (freeCount > 0)
```
eriove: Code coverage suggests that `freeCount` is never larger than 0. Experimented with test that would solve cover this path but didn't succeed.

eriove:
Also found a class named BfgsSolver. It seems quite redundant at the moment. Should I just remove it? It probably belongs to the (not yet created) facade class in root namespace

cdrnet:
Thanks a lot for moving forward with this!

Regarding test cases, we have a few more common test functions defined in [TestFunctions](https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics/SpecialFunctions/TestFunctions.cs), maybe we could add their gradients there and extend the tests to cover some of them as well?

eriove:
I will try to add the gradients and run tests with these functions to. I'll also try to find my old optimization textbook from university to see if there are any line search examples that might help to increase the code coverage.

cdrnet:
I think we can merge this already anyway?

eriove:
Sure. I'll create a new pull request when I find the time to write more tests.

PR #407, opened 2016-06-20
fixed documentation of Vector.MaximumIndex()

wojcech:
"returns Index of absolute maximum -> returns Index of Maximum"


cdrnet:
Good catch, thanks for the fix!

PR #409, opened 2016-06-24
Multiplying sparse with dense matrices with DenseColumnMajorMatrixStorage

arthurvb:
Directly address underlying data-array for multiplying sparse with dense matrices with DenseColumnMajorMatrixStorage


cdrnet:
Great, thanks!

In the end we will also need the same for complex matrices (an unfortunate consequence of the current design). Would you be willing to provide them as well as part of this PR? Otherwise I can add them myself along the same lines.

arthurvb:
No problem, just added commit for complex types,
Thank you for the great library!

Arthur

cdrnet:
Thanks again!

PR #410, opened 2016-06-25
fixed generation of ChiSquared Samples

florianwechsung:
I noticed a bug in the generation of ChiSquared samples. Probably left from copy-pasting code from Chi.cs


cdrnet:
Great catch, thanks for reporting and fixing this!

cdrnet:
Yes, this looks like a copy-paste error indeed.

PR #415, opened 2016-07-05
Correlation

StratThinker:
There was an incorrect formula in the weighted correlation method.  I have corrected it, I have added a unit test which would fail in the previous case but it passes now.


cdrnet:
Yes, this looks correct to me. Thanks for reporting this and for the fix!

PR #416, opened 2016-07-06
R squared

StratThinker:
I added a method "CoefficentOfDetermination" which calculates the R^2 value.  Currently the method "RSquared" calculates r^2 (the square of the correlation between the model and data), now r^2 is equal to R^2 when the model was created via least squares linear regression.  However in other cases r^2 is not equal to R^2.  

I was not sure whether to replace the old implementation of "RSquared" or write a new method. According to [Wikipedia](https://en.wikipedia.org/wiki/Coefficient_of_determination#As_squared_correlation_coefficient) some people use r^2 to determine how good a least squares line fit would have been; thus rewriting "RSquared" might be a breaking change; thus I opted, for now, to write a new method (although maybe we should add a "rSquared" method and mark "RSquared" as obsolete(??)).

I also added three unit tests.

### DIFF
--- a/src/Numerics/GoodnessOfFit.cs
+++ b/src/Numerics/GoodnessOfFit.cs
@@ -48,13 +48,13 @@ namespace MathNet.Numerics
         }
 
         /// <summary>
-        /// Calculated the R-Squared value, also known as coefficient of determination,
-        /// given modelled and observed values
+        /// Calculates the R-Squared value, also known as coefficient of determination,
+        /// given some modelled and observed values.
         /// </summary>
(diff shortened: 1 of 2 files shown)


cdrnet:
Thanks for bringing this up.

The current implementation could be considered as a bug, if the documentation had not explicitly mentioned the correlation.

Although closer to common mathematical notation, having both `rSquared` and `RSquared` is not an option form an API viewpoint for me. What about `RSquared` and `CorrelationSquared` as "desired" naming?

cdrnet:
Finally merged. Thanks again!

PR #417, opened 2016-07-09
Create Xamarin iOS specific project

praeclarum:
This adds a Xamarin.iOS project (non PCL). This is needed by my app [Continuous](http://continuous.codes), but I hope it will also serve as a base for extending the library to use Apple's native Accelerate framework (BLAS and LAPACK).


cdrnet:
Could it work to use globbing here like we do with the other derived projects, e.g. [here](https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics/Numerics-Net35.csproj#L93)? Mainly to make sure we cannot forget to update it when files are being added. This is assuming any development happens on the primary project.

praeclarum:
Ah, it started out globbed but Xamarin Studio trickily changed it on me. Fixing now...

cdrnet:
I assume the output of this project would have to end up in the NuGet package in the end?

praeclarum:
In the end yes, but I don't see any reason now until it actually has the native providers working.

cdrnet:
NB: Would be great indeed if we could support Apple's native Accelerate framework as linear algebra native provider!

praeclarum:
I added the F# project too but had to change to use the bigint `( - )` operator. Xamarin for some reason doesn't include `bigint.Negate`.

cdrnet:
Thanks!

Issue #424, opened 2016-07-29
Multiplying tall matrices is extremely slow


florianwechsung:
I have noticed that multiplying a tall matrix with a small matrix is extremely slow and does not scale linearly in the tallness of the matrix. 

Example:
Let n>>d, T\in R^{n\times d} and S\in R^{d\times d}, then the time to calculate the product TS should scale linearly in n.

Code: 

        [Test]
        public void MatrixMultiplyTest()
        {
            var ns = new List<int> {5000, 10000, 20000, 40000};
            var d = 10;
            var small = DenseMatrix.CreateRandom(d, d, new ContinuousUniform());
            foreach (var n in ns)
            {
                var tall = DenseMatrix.CreateRandom(n, d, new ContinuousUniform());
                var sw1 = new System.Diagnostics.Stopwatch();
                sw1.Start();
                var prod1 = tall.Multiply(small);
                sw1.Stop();
                var sw2 = new System.Diagnostics.Stopwatch();
                sw2.Start();
                var prod2 = DenseMatrix.OfColumnVectors(tall.EnumerateRows().Select(row => small.Transpose().Multiply(row))).Transpose();
                sw2.Stop();
                Console.WriteLine("n = " + n + ", MathNet: " + sw1.ElapsedMilliseconds/1000.0 + "s, Naive: " + sw2.ElapsedMilliseconds/1000.0 + "s.") ;
                Assert.That(prod1.ToArray(), Is.EquivalentTo(prod2.ToArray()));
            }
        }

Output: 

        n = 5000, MathNet: 0,208s, Naive: 0,026s.
        n = 10000, MathNet: 1,13s, Naive: 0,01s.
        n = 20000, MathNet: 7,714s, Naive: 0,021s.
        n = 40000, MathNet: 66,842s, Naive: 0,064s.

I am especially confused how enumerating over the rows can be so much faster than the normal matrix multiplication and why the runtime does not scale linearly.

Aixile:
Math.Net use divide and conquer to do matrix multiplication by default.

In function CacheObliviousMatrixMultiply, the structure is like

```
static void CacheObliviousMatrixMultiply(Transpose transposeA, Transpose transposeB, float alpha, float[] matrixA, int shiftArow, int shiftAcol, float[] matrixB, int shiftBrow, int shiftBcol, float[] result, int shiftCrow, int shiftCcol, int m, int n, int k, int constM, int constN, int constK, bool first)
{
[...]

PR #426, opened 2016-08-18
Fix recursion issue of CacheObliviousMatrixMultiply

Aixile:
See https://github.com/mathnet/mathnet-numerics/issues/424


cdrnet:
Thanks!

Issue #427, opened 2016-08-23
Result of RandomSource.Next(x, x+1) Not in Range [x, x+1[


Genmutant:
Using MathNet.Numerics 3.13.0:

When I use the code `RandomSource.Next(x, x+1)`, I don't get the expected value of x back, but something (random) out of range.

Testcode:
```
RandomSource source = new Xorshift();
var next = source.Next(9, 10);
next.Should().Be(9); // Expected value to be 9, but found -2037414154.
```

Maybe there should be a fastpath added, to just return the minimum element without a random call?

Genmutant:
The problem seems to be, that `DoSampleInt32WithNBits` is called with the parameter `bitCount=0`, and it tries to shift with an invalid amount of 32 bits:
```
// the least significant N bits with independend uniform distribution and the remaining bits zero
            uint uintN = uint32 >> (32 - bitCount);
```
Which leaves `uintN` the same as `uint32`.

cdrnet:
Thanks for catching and reporting this!

cdrnet:
Merged to master

PR #428, opened 2016-08-23
Fix for Issue #427

Genmutant:
I added two special cases:
* when 0 bits should be sampled, alsways return 0
* when an int should be sampled with an exclusive maximum of 1, always return 0

This should fix Issue #427 .
I also added a new test, if the boundaries are respected when calling Next(). I'm not  completely sure it's in the right place though.


cdrnet:
Thanks!

PR #429, opened 2016-08-25
Added segment index caching for more efficient interpolation of large…

AdsRSherman:
I added a cache to linear, cubic, and quadratic spline interpolators to prevent unnecessary binary searches.  

Very simply, it won't redo a binary search if the last returned index is valid for the current query.

### DIFF
--- a/src/Numerics/Interpolation/CubicSpline.cs
+++ b/src/Numerics/Interpolation/CubicSpline.cs
@@ -40,6 +40,9 @@ namespace MathNet.Numerics.Interpolation
     /// <remarks>Supports both differentiation and integration.</remarks>
     public class CubicSpline : IInterpolation
     {
+        private int _cachedLeftSegmentIndex;
+        private double _cachedLeftSegment;
+        private double _cachedRightSegment;
         readonly double[] _x;
         readonly double[] _c0;
         readonly double[] _c1;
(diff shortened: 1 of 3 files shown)


cdrnet:
Thanks for proposing this change! It does indeed likely improve performance on typical use cases.

Unfortunately it does seem like this could break in multi-threaded use cases though, i.e. when the resulting interpolation is used concurrently from multiple threads.

AdsRSherman:
Gotcha.  I can change the caches to ThreadLocal caches if that would help.  

Also, I have n-degree lagrange and hermite polynomial interpolators that implement IInterpolation if you're interested.  The algorithms were borrowed from JPL's SPICE toolkit.

AdsRSherman:
@cdrnet Added thread-local caches.  In my testing this substantially improves performance, especially when abscissa are sparse and queries are dense.

AdsRSherman:
In other news, I'm not rightfully sure why those checks would have failed.

AdsRSherman:
@cdrnet 

I have 3 other polynomial interpolations implemented: [lagrange](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/lgrint.html), [hermite](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/hrmint.html), and [chebychev](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/chbint.html).

Is there interest in my adding those to the pr?

PR #435, opened 2016-09-24
Better tests for optimization

scottstephens:
Commit messages have details.

Areas for comment:
- I modified LazyObjectiveFunctionBase to allow subclasses to reuse the memory. I didn't design this class, and I fear I may have done this clumsily. Would be interested in input on whether or not this is the best design to meet my goal of memory reuse.
- Currently some of the tests that work with BFGS and BFGS-B are failing for conjugate gradient, Newton, and Nelder-Mead minimizers (but not the same set for all). I didn't do a deep dive, but I suspect these are failings of the algorithms rather than the implementation. I'm not sure how we should deal with cases like this; for the time being, I marked them as ignored.


eriove:
I'll try to find some time tomorrow to look at the failing tests.

eriove:
I'm currently cross checking all the equations, it would be great if you could add references to the equations for the test functions. It's quite easy to find them by Google-ing but they tend to be in other forms so it would ease the cross checking to have the same source.

scottstephens:
All of the test functions are from 

Testing Unconstrained Optimization Software
Jorge J. Moré, Burton S. Garbow, Kenneth E. Hillstrom
ACM Transactions on Mathematical Software, Vol 7, No. 1, March 1981, Pages 17-41.

Full text available here:
http://www.caam.rice.edu/~yzhang/caam454/nls/MGH.pdf

Agree this info should be added to the code somewhere.

cdrnet:
Thanks a lot!

PR #439, opened 2016-10-12
Optimization: add finite difference objective function adapter

scottstephens:
Added objective function adapter that can be used to calculate gradients by forward finite difference if an analytical gradient is not available.


cdrnet:
Thanks!

PR #445, opened 2016-10-28
Fix multivariate Runge-Kutta ODE solver typo

Ksero:
The multiplication by the time-step was missing in this method (but not the others)


cdrnet:
Yes, this indeed looks like a typo. Thanks a lot for figuring it out and providing the fix!

PR #450, opened 2016-11-19
Any interests in element wise operations on vector/matrix with F#?

albertp007:
Discussion thread is here:  http://discuss.mathdotnet.com/t/any-interests-in-element-wise-operations-on-vector-matrix-with-f/367

### DIFF
--- a/src/Numerics/LinearAlgebra/Complex/Matrix.cs
+++ b/src/Numerics/LinearAlgebra/Complex/Matrix.cs
@@ -461,7 +461,7 @@ namespace MathNet.Numerics.LinearAlgebra.Complex
 
         protected override void DoPointwiseAbs(Matrix<Complex> result)
         {
-            Map(x => (Complex)Complex.Abs(x), result, Zeros.Include);
+            Map(x => (Complex)Complex.Abs(x), result, Zeros.AllowSkip);
         }
         protected override void DoPointwiseAcos(Matrix<Complex> result)
         {
@@ -469,11 +469,11 @@ namespace MathNet.Numerics.LinearAlgebra.Complex
         }
         protected override void DoPointwiseAsin(Matrix<Complex> result)
         {
-            Map(Complex.Asin, result, Zeros.Include);
+            Map(Complex.Asin, result, Zeros.AllowSkip);
         }
         protected override void DoPointwiseAtan(Matrix<Complex> result)
         {
-            Map(Complex.Atan, result, Zeros.Include);
+            Map(Complex.Atan, result, Zeros.AllowSkip);
         }
         protected override void DoPointwiseAtan2(Matrix<Complex> other, Matrix<Complex> result)
         {
@@ -509,23 +509,23 @@ namespace MathNet.Numerics.LinearAlgebra.Complex
         }
         protected override void DoPointwiseSin(Matrix<Complex> result)
         {
-            Map(Complex.Sin, result, Zeros.Include);
+            Map(Complex.Sin, result, Zeros.AllowSkip);
         }
         protected override void DoPointwiseSinh(Matrix<Complex> result)
         {
-            Map(Complex.Sinh, result, Zeros.Include);
+            Map(Complex.Sinh, result, Zeros.AllowSkip);
         }
         protected override void DoPointwiseSqrt(Matrix<Complex> result)
         {
-            Map(Complex.Sqrt, result, Zeros.Include);
+            Map(Complex.Sqrt, result, Zeros.AllowSkip);
         }
         protected override void DoPointwiseTan(Matrix<Complex> result)
         {
-            Map(Complex.Tan, result, Zeros.Include);
+            Map(Complex.Tan, result, Zeros.AllowSkip);
         }
         protected override void DoPointwiseTanh(Matrix<Complex> result)
         {
-            Map(Complex.Tanh, result, Zeros.Include);
+            Map(Complex.Tanh, result, Zeros.AllowSkip);
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/Numerics/LinearAlgebra/Matrix.Arithmetic.cs
```diff
@@ -1512,38 +1529,77 @@ public void PointwiseRemainder(Matrix<T> divisor, Matrix<T> result)
...
         /// <summary>
-        /// Pointwise applies the exponent function to each value.
+        ///  
         /// </summary>
-        /// <param name="result">The matrix to store the result.</param>
-        /// <exception cref="ArgumentException">If this matrix and <paramref name="result"/> are not the same size.</exception>
-        public void PointwiseExp(Matrix<T> result)
+        /// <param name="f"></param>
+        /// <returns></returns>
+        protected void PointwiseUnary(Action<Matrix<T>> f, Matrix<T> result)
```
cdrnet: I wonder whether we actually need these helpers, since we already do range checking within the Map function. Or we could use the unchecked Map implementations. Seems I need to spend some thoughts on the architecture here - but that should not block this PR.

src/Numerics/LinearAlgebra/Matrix.Arithmetic.cs
```diff
@@ -1553,12 +1609,315 @@ public Matrix<T> PointwiseLog()
...
-                throw DimensionsDontMatch<ArgumentException>(this, result);
-            }
+            PointwiseUnary(DoPointwiseLog, result);
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <returns></returns>
+        public Matrix<T> PointwiseAbs()
```
cdrnet: We'll need to have proper documentation on all public members in the end.

src/Numerics/LinearAlgebra/Matrix.Operators.cs
```diff
@@ -305,5 +305,90 @@ public static Matrix<T> op_DotHat(Matrix<T> matrix, T exponent)
         {
             return matrix.PointwisePower(exponent);
         }
+
+        public static Matrix<T> Log(Matrix<T> x)
```
cdrnet: These will also need proper documentation.

src/Numerics/LinearAlgebra/Vector.Arithmetic.cs
```diff
@@ -957,53 +975,440 @@ public void PointwiseRemainder(Vector<T> divisor, Vector<T> result)
...
         /// <summary>
-        /// Pointwise applies the exponent function to each value.
+        /// 
         /// </summary>
-        /// <param name="result">The vector to store the result.</param>
-        /// <exception cref="ArgumentException">If this vector and <paramref name="result"/> are not the same size.</exception>
-        public void PointwiseExp(Vector<T> result)
+        /// <param name="f"></param>
+        /// <param name="result"></param>
+        protected void PointwiseUnary(Action<Vector<T>> f, Vector<T> result)
```
cdrnet: Please avoid adding empty documentation skeletons, that makes it hard to spot whether we're still lacking some proper documentation.

src/Numerics/LinearAlgebra/Complex/Matrix.cs
```diff
@@ -459,6 +459,75 @@ protected override void DoPointwiseLog(Matrix<Complex> result)
             Map(Complex.Log, result, Zeros.Include);
         }
 
+        protected override void DoPointwiseAbs(Matrix<Complex> result)
+        {
+            Map(x => (Complex)Complex.Abs(x), result, Zeros.Include);
```
albertp007: Ah, got it, thanks!  Will fix it for abs, acos, asin, atan, ceiling, floor, round, sin, sinh, sqrt, tan and tanh

src/Numerics/LinearAlgebra/Matrix.Arithmetic.cs
albertp007: These are really because I saw the checks done in all the other pointwise functions before this pull request.  Since I am adding some 20 more of them, copy and pasting the checks is something I really want to avoid.  True, I did not try to refactor the existing ones though.  Yeah, please have a think and we can decide what to do here.

src/Numerics/LinearAlgebra/Vector.Arithmetic.cs
albertp007: True, will fix and avoid going forward.

src/Numerics/LinearAlgebra/Matrix.Operators.cs
albertp007: Yup, will fix shortly, thanks!

src/Numerics/LinearAlgebra/Matrix.Arithmetic.cs
cdrnet: Thanks. It's ok for now, but I'll think about it.

src/Numerics/LinearAlgebra/Matrix.Arithmetic.cs
cdrnet: Thanks. Yes, It's quite tedious; sorry for all the duplication (this would be less an issue with code gen, but we move away from that a long time ago for good reasons).

cdrnet:
Merged, thanks a lot!

I hope you do not mind that I let GitHub squash the commits together.

albertp007:
Thank you!  Don't mind at all, would have been what I wanted to do anyways instead of having so many incremental commits.  Glad to be able to contribute!

PR #459, opened 2016-12-10
Minor improvement to "Getting started".

davidfgarcia:
Adding some .NET3.5/Win10 troubleshooting steps that I had to follow in order to get to a clean build, in case they help someone else.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,4 @@ docs/content/Contributing.md
 docs/content/Contributors.md
 docs/content/ReleaseNotes.md
 docs/content/ReleaseNotes-*.md
+/mathnet-numerics/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,11 @@ We use the [Fork & Pull Model](https://help.github.com/articles/using-pull-reque
 **C# Solutions, Projects and Files**  
 We have two kind of C# projects: primary (*Numerics.csproj, UnitTests.csproj*) and secondary (*Numerics-xy.csproj, UnitTests-xy.csproj*). The primary ones are the common VisualStudio project files you usually work with. The secondary projects on the other hand are not intended to be modified and include all files automatically. Whenever you need to add, remove or move a file, please do so in the primary projects only. In most cases we recommend to work with the `MathNet.Numerics.sln` solution which only includes primary projects anyway - except when working on and testing portability/compatibility.
 
+If you are not familiar with F#/paket, follow these steps in order to avoid F#-related errors:
+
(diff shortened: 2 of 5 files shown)


cdrnet:
Did you actually need .Net 3.5 builds? The normal .Net 4 build should not need anything related to 3.5 (if it does, we should fix it).

davidfgarcia:
That was my experience, yes, I think the culprit was project “TestData”.

Once .NET 3.5 was installed and enabled on my machine (Win10, VS2015), I got a fully clean build.

davidfgarcia:
Sure!

I’m very new to GitHub; I saw those two files in the changeset but they were not marked as “committed”, so I didn’t know what GH was going to do about them.

I’ll remove them and re-submit the pull request.

Thanks a lot!

davidfgarcia:
Based on your other email, I guess the “Appendix” I added will have to be removed once the dependency on .NET 3.5 is removed? Does it make sense to add if for now, or should we make sure the solution builds without .NET 3.5 instead?

From: David Figueiras Garcia
Sent: Saturday, December 10, 2016 10:29 AM
To: 'mathnet/mathnet-numerics' <reply@reply.github.com>; mathnet/mathnet-numerics <mathnet-numerics@noreply.github.com>
Subject: RE: [mathnet/mathnet-numerics] Minor improvement to "Getting started". (#459)

Sure!

I’m very new to GitHub; I saw those two files in the changeset but they were not marked as “committed”, so I didn’t know what GH was going to do about them.

I’ll remove them and re-submit the pull request.

Thanks a lot!

cdrnet:
Ah yes - this has indeed slipped in. We need TestData for all builds so I thought building it for .Net 3.5 should make it work without having to create 10 different project files for each target, but seems we cannot take that shortcut.

cdrnet:
I think we should fix the solution to only require 3.5 if we actually want to build the 3.5 variant. The appendix info might still be useful for this, since there may be a need to build 3.5 on Windows 10 (e.g. when doing a full release), but maybe we could move it to [Build](http://numerics.mathdotnet.com/Build.html) instead of GettingStarted.

davidfgarcia:
Replacing with new pull request.

davidfgarcia:
Makes sense, I replaced the old PR with a new one (couldn’t figure out how to modify the existing one):

https://github.com/mathnet/mathnet-numerics/pull/460

Pull requests, commits and etc. will get a bit cleaner as I get more familiar with GitHub I hope. ☺

Thanks a lot, -david

davidfgarcia:
Opened new issue to track the work suggested below:

https://github.com/mathnet/mathnet-numerics/issues/461

Issue #412, opened 2016-06-27
Standard error, T-stat, p value, F-stat for linear regression. 


fantested:
Can anyone please implement these?

bwang22:
I second this

PR #464, opened 2016-12-19
Added code to calculated Standard Error of the Regression for a linear…

ardave:
… model.

Feedback welcome.  If this is suitable, I can continue on with the rest of the parts of Issue #412 .   

If not, please give feedback and I can rework.

### DIFF
--- a/src/Numerics/GoodnessOfFit.cs
+++ b/src/Numerics/GoodnessOfFit.cs
@@ -66,7 +66,21 @@ namespace MathNet.Numerics
         /// <param name="modelledValues">The modelled/predicted values</param>
         /// <param name="observedValues">The observed/actual values</param>
         /// <returns>The Standard Error of the regression</returns>
-        public static double StandardError(IEnumerable<double> modelledValues, IEnumerable<double> observedValues)
+        public static double PopulationStandardError(IEnumerable<double> modelledValues, IEnumerable<double> observedValues)
+        {
+            return SampleStandardError(modelledValues, observedValues, 0);
+        }
+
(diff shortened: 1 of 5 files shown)


cdrnet:
Thanks for putting this together, including the test!

I do wonder about the right choice for the denominator. In `Statistics` we typically have two variants, one for cases where the provided data is the full population (e.g. `Statistics.PopulationVariance`) and one for cases where the data is a sample only (e.g. `Statistics.Variance`). The former typically has `N` as denominator, the latter `N-1`. There is a similar note in the linked page, arguing that in the sampling case, the denominator shall be `N-k` where `k` is the order (degree of freedom) of the regression. I wonder whether we should provide two variants as well, or have an additional `degreesOfFreedom` or `order` argument to control the normalization?

ardave:
I can certainly do that, and submit the changes in a new PR.

ardave:
See if I interpreted the requirements correctly with this second commit.

I can continue trying to rebase/squash the commits into 1 if it's important - git is acting up on my machine and feeding gibberish into my rebase-todo file right now.

Also, I can change the method signatures to require arrays rather than ienumerables; doing so would allow me to guard more proactively against the `degreesOfFreedom > numberOfSamples` case, at the top of the method, by relying on the array.length properties rather than materializing the sequences.  For now, I've just tried to mimic the style of the existing code that is closest to the new code that I have added.

cdrnet:
Merged - thanks again!

ardave:
Cool.  I'll continue on now with the more parts of this Issue.

PR #473, opened 2017-01-21
Support netstandard 1.1

iamcarbon:
- Introduces new compile time symbols for NETSTANDARD1_0, NETSTANDARD1_3,  NOCLONABLE and NOSERIALIZATION.

- Adds a project.json file, loaded through "MathNet.Numerics.Core.sln" targeting NETSTANDARD1.0, NETSTANDARD1.3, NET35, and NET451. The desktop assemblies remain dependency free. Both NETSTANDARD1.3 & NET451 provide support for the native providers plugins.

- Includes a build script to generate a nuget package (core/build.cmd)


iamcarbon:
Added partial support for native providers in NETSTANDARD1.3 although we may want to wait until NETSTANDARD 2.0 to enable this for full parity.

src/Numerics/Providers/Compatibility.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    using System;
+
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
+    internal class SuppressUnmanagedCodeSecurityAttribute : Attribute
+    {
+    }  
+}
+
+#endif
```
cdrnet: We should add this file also to Numerics.csproj (which should include all cs files as it is the primary project which is manually maintained).

src/Numerics/project.json
```diff
@@ -0,0 +1,97 @@
...
+  "authors": [ "Christoph Ruegg", "Marcus Cuda", "Jurgen Van Gael" ],
+
+  "version": "3.17.0",
+  "dependencies": {
+
+  },
+
+  "frameworks": {
+    "netstandard1.0": {
+      "imports": "dnxcore50",
```
cdrnet: Why would .Net Standard import .Net Core? Isn't the `dnx` stuff long deprecated? 😕

src/Numerics/project.json
```diff
@@ -0,0 +1,97 @@
+﻿{
+  "name": "MathNet.Numerics.Core",
```
cdrnet: If possible we should avoid to split the ecosystem even further. This PR is adding .Net Standard support exactly to avoid splitting between .Net and .Net Core unnecessarily. I think we should try to have just one package supporting everything, without the Core suffix.

src/Numerics/project.json
```diff
@@ -0,0 +1,97 @@
...
+        "System.Runtime.Serialization": "3.0.0.0"
+      },
+
+      "buildOptions": {
+        "define": [ "NET35", "NOSYSNUMERICS" ],
+        "optimize": true
+      }
+    },
+
+    "net451": {
```
cdrnet: We currently do not yet explicitly build for .Net 4.5.1, but we have a build for .Net 4.0 and 4 PCL builds. Is this supported by this approach as well? (No matter if not, not worth investing time, just wondering)

src/Numerics/project.json
```diff
@@ -0,0 +1,97 @@
+﻿{
```
cdrnet: The mere presence of a project.json file used to cause havoc on the other projects. I didn't run into any issues yet though, neither in the build nor in VS. Has this been fixed then somehow? That would be great, this was a major blocker up to now.

core/build.cmd
```diff
@@ -0,0 +1,2 @@
+@ECHO OFF
+PowerShell -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = '';& '%~dp0build.ps1' %*"
```
cdrnet: This will work fine for experiments and first previews - to make this all working. But in the long run this needs to be integrated into our full build system. We do (and need to do) much more than just build and package binaries into a NuGet nupkg.

cdrnet:
Thanks a lot for moving forward with this! Greatly appreciated!

src/Numerics/Mathnet-Core.xproj
```diff
@@ -0,0 +1,21 @@
...
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">.\obj</BaseIntermediateOutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">.\bin\</OutputPath>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <SchemaVersion>2.0</SchemaVersion>
+  </PropertyGroup>
+  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.targets" Condition="'$(VSToolsPath)' != ''" />
+</Project>
```
iamcarbon: This is developed against preview 2 tooling while we wait on VS2017 to be released. I just tested the migrating to the new csproj format and everything works as expected including support for all the existing portable & desktop targets.  It will be super easy to map all this over when we're ready.

src/Numerics/project.json
iamcarbon: Have used this approach on some other projects and haven't had any issues yet. I think this will work OK until we can migrate to VS2017 after RTM and merge everything into a unified csproj file.

iamcarbon:
During testing, I realized .NETSTANDARD 1.0 doesn't support concurrent collections and the parallel  task library. Bumped this to .NETSTANDARD 1.1 which drops support for Silverlight on Phone.

src/Numerics/Providers/Common/NativeProviderLoader.cs
```diff
@@ -72,13 +72,30 @@ static bool IsUnix
...
                 return p == PlatformID.Unix || p == PlatformID.MacOSX;
+
+#endif
             }
         }
 
         static string EvaluateArchitectureKey()
         {
+#if NETSTANDARD1_3
+            switch (RuntimeInformation.OSArchitecture)
```
iamcarbon: Need to check the process. I think sizeof(IntPtr) should work -- testing now.

src/Numerics/project.json
iamcarbon: Changed back to MathNet.Numerics.

src/Numerics/project.json
iamcarbon: Yep, no longer needed. Removed.

src/Numerics/project.json
iamcarbon: Changed to net40. 

We may want to consider targeting modern frameworks as more optimizations come aboard (i.e. Split(char), Array.Empty<>, etc) -- but better to keep consistent for now.

src/Numerics/Random/RandomSeed.cs
```diff
@@ -1,15 +1,16 @@
-﻿using System;
-
-namespace MathNet.Numerics.Random
+﻿namespace MathNet.Numerics.Random
 {
+    using System;
+    using System.Security.Cryptography;
```
iamcarbon: Shouldn't have changed this. Updated.

src/Numerics/Exceptions.cs
```diff
@@ -20,7 +20,7 @@ public NonConvergenceException(string message) : base(message)
         public NonConvergenceException(string message, Exception innerException) : base(message, innerException)
         {
         }
-#if !PORTABLE
+#if !PORTABLE && !NOSERIALIZATION
```
iamcarbon: We should consider removing the PORTABLE symbol all together in favor of feature flags.

cdrnet: Just for the record, the symbols we currently use are documented [here](https://github.com/mathnet/mathnet-numerics/blob/master/docs/content/Compatibility.md) (rendered [here](https://numerics.mathdotnet.com/Compatibility.html#Platform-Discrepancies))

src/Numerics/Providers/Compatibility.cs
iamcarbon: Merged this in with the main compatibility file to keep things simple.

src/Numerics/Providers/Common/NativeProviderLoader.cs
iamcarbon: Correctly checks the process bitness now.

src/Numerics/LinearAlgebra/Matrix.BCL.cs
```diff
@@ -76,7 +76,7 @@ public override int GetHashCode()
             return Storage.GetHashCode();
         }
 
-#if !PORTABLE
+#if !PORTABLE && !NOCLONABLE
```
iamcarbon: Updated to NOCLONEABLE

iamcarbon:
@cdrnet Thanks for pointing me to the symbols/ platform guide. I'll start the work on replacing the PORTABLE flags with feature flags.

After this work is completed, please let me know your thoughts on integrating this to master before we begin the build changes. There's a discussion there on whether we want to wait ~3 months for the RTM of msbuild15 / VS2017 (supporting the new csproj format) -- or get things running on preview2 tooling.

iamcarbon:
@cdrnet Quick ping on thoughts on upgrading the project to VS2017 and dropping support for portable profiles for .NETStandard. Can start this work --

lennoncork:
Is anyone working on a 2017 build for multiple targets? We have a need to target netstandard, and could do part of the work to port to a 2017 build. Just want to make sure that it's not already in progress.

cdrnet:
Related: #499

Huqin-China:
since netcore2.0 support most netframework dll's,seems no need to develop a netstandard version now?

iamcarbon:
Closing this now that .NETSTANDARD support has landed in another PR! HOORAY!

PR #482, opened 2017-03-24
Handle combining empty running statistics

TrexinanF14:
This prevents most of the fields from being set to NaN when merging an empty instance.


cdrnet:
Thanks!

PR #486, opened 2017-04-15
Fix docs for DoPointwiseMultiply

darthdeus:


cdrnet:
Good catch, thanks a lot!

PR #497, opened 2017-05-28
BetaRegularized Fix - Issue 491

eliasaj92:
Changed the number of iterations for the loop from 140 to 50,000 in the BetaRegularized method. This brought the result of 
```
MathNet.Numerics.SpecialFunctions.BetaRegularized(6985172036240.918, 5029323866093.4609, 0.581395348837209);
``` 
to 0.5101 which seem to agree with the results given by R(0.5) and Mathematica(0.495).


cdrnet:
We'll need a better way at some point, but I'll accept this for now. Thanks for looking into this and for providing the PR!

PR #506, opened 2017-08-14
Adjusted Neldermead

MMAvanEnkhuizen:
Adjusted the algorithm such that after a reflection it will also try to expand or shrink (as said in the documentation). Now it also works for the 1 dimensional issue I stated (see Issue 501)


MMAvanEnkhuizen:
Looks like there is a minor problem with one of the tests. Will look to fix this in the upcoming weeks

PR #511, opened 2017-08-29
initial L-BFGS implementation

florianwechsung:
First draft for a L-BFGS implementation, see [here](https://discuss.mathdotnet.com/t/l-bfgs-optimization-algorithm/521).

I think it would be nice if the user could add a callback to observe the status of the optimization at each step. Is something like this done in Math.NET somewhere, so that I have a place to start?

### DIFF
--- a/src/Numerics/Numerics.csproj
+++ b/src/Numerics/Numerics.csproj
@@ -116,6 +116,8 @@
     <Compile Include="LinearRegression\Options.cs" />
     <Compile Include="OdeSolvers\AdamsBashforth.cs" />
     <Compile Include="OdeSolvers\RungeKutta.cs" />
+    <Compile Include="Optimization\MinimizerBase.cs" />
+    <Compile Include="Optimization\LimitedMemoryBfgsMinimizer.cs" />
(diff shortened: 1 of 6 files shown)


cdrnet:
We have isolated the iterators in the linear iterative solvers (e.g. [GpBiCg](https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics/LinearAlgebra/Double/Solvers/GpBiCg.cs)), where users can not only observe the steps but also control them arbitrarily.

florianwechsung:
I'm looking into fixing the errors on AppVeyor and Travis, but I'm a bit lost. I don't see which unit test the error message is referring to, instead it says that Fake has a caching issue. The tests also pass if I run them locally from visual studio. 

Any chance you could point me in the right direction?

cdrnet:
It seems both failures are unrelated to your changes: Travis seems to have an issue with the particular combination of Mono and NUnit at the moment, and the one failing unit test in AppVeyor cannot possibly be related to your changes (MathNet.Numerics.UnitTests.DistanceTests.Euclidean).

cdrnet:
I've finally merged in the change as is manually (to resolve the conflict).
Thanks a lot!

PR #536, opened 2018-01-04
Initial Jacobian of Broyden method

AappoPulkkinen:
I noticed that Broyden method fails to find the root in some cases. To my knowledge, the reason for this is that the step size for calculating the Jacobian matrix can become smaller than the machine epsilon. This causes the Jacobian matrix to be singular.

Two changes were made to the Broyden method to fix the issue:
1. Added the possibility to give the step size for calculating the approximate Jacobian as a parameter. Optional parameter is added to method FindRoot and alternative method TryFindRootWithJacobianStep is added.
2. Changed the formula for step size to be h = (1.0 + Abs(x))*stepSize instead of Abs(x)*stepSize. This fixes the problem with too small step size and is in line with the implementation in NumericalDerivative class.

A unit test demonstrating the failure is added. Also the unit tests that were broken by the change are fixed.

### DIFF
--- a/src/FSharp/FindRoots.fs
+++ b/src/FSharp/FindRoots.fs
@@ -70,6 +70,11 @@ module FindRoots =
         | true, root -> Some root
         | false, _ -> None
 
+    let broydenWithJacobianStep jacobianStepSize maxIterations accuracy guess (f:float[]->float[]) =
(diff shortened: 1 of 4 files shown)


cdrnet:
Thanks a lot for the analysis and the fix. The proposed change looks reasonable to me.

Technically the addition of the optional parameters is breaking, but this is ok for v4 (but if we ever need to backport it to v3 we'd have to skip if or use an overload instead).

Issue #539, opened 2018-01-09
Evaluation of Coth(z)


diluculo:
In the `MathNet.Symbolics`, I have found `Coth(z)` function gives incorrect value. The problem is the wrong sign of imaginary part:

> coth(1 + Expression.I) |> Approximate.approximate;;
val it : Expression = 0.868014142895925 + 0.217621561854403*j

> cosh(1 + Expression.I) / sinh(1 + Expression.I) |> Approximate.approximate;;
val it : Expression = 0.868014142895925 - 0.217621561854403*j

Actually it is a bug in `Coth(z)` function of `MathNet.Numerics`. Its last line should be updated as follows:

```
public static Complex Coth(this Complex value)
{
      ...
      //return new Complex(sinhr * Cosh(value.Real) / denom, sini * Cos(value.Imaginary) / denom);
      return new Complex(sinhr * Cosh(value.Real) / denom, - sini * Cos(value.Imaginary) / denom);
}
```

cdrnet:
For reference, according to Mathematica:

![coth](https://user-images.githubusercontent.com/47147/34727892-8d25dc5c-f558-11e7-9a6d-f7568c32e9b7.png)

diluculo:
Yes, so we can simply prove that imaginary part of `cosh(x + j y)` can be written as 
`sin(2*y)/(cos(2*y) - cosh(2*x)) = - sin(y)*cos(y)/(sin(y)^2 + sinh(x)^2)` 

by using
- `sin(2*y) = 2*sin(y)*cos(y)`
- `cos(2*y) = cos(y)^2 - sin(y)^2 = 1 - 2*sin(y)^2`
- `cosh(2*x) = cosh(x)^2 + sinh(x)^2 = 1 + 2*sinh(x)^2`

cdrnet:
Excellent, thanks! I'll provide a fix.

cdrnet:
Fixed in master. Thanks again for reporting.

To be released as part of v4.0.0-beta03. Let me know if there is a need to backport to v3 as well.

diluculo:
@cdrnet Many thanks for very quick update.

Just now, I find another problem: `Coth(500 + 500 j)` gives `0`. It should return `1`. It happens in Tanh(z), too. Inside the code, you can watch `sinhr = 7.012E+216 != oo`, so `denom = oo`. Please check if `denom` or `sinhr * sinhr` is `Infinity` as follows.:

```
[...]

PR #540, opened 2018-01-16
Updated some (hyperbolic) trig functions

diluculo:
PR for #539

### DIFF
--- a/src/Numerics/Trigonometry.cs
+++ b/src/Numerics/Trigonometry.cs
@@ -447,9 +447,20 @@ namespace MathNet.Numerics
                 return new Complex(Sinh(value.Real), 0.0);
             }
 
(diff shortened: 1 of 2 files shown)


diluculo:
I missed a case of infinite `denom`. I will be back. Sorry.

diluculo:
Done.

To avoid overflow, if `x > asinh(max)/4 = 177.62`, the approximation of `sinh^2(x) ~= exp(-2|x|)/4` is applied. This limit value was from Kahan's paper. Please see the updated pdf file in #539.

For reference, in the first commit, `x > asinh(sqrt(max))/4 = 355.58` was used and I found a overflow of `denom` in 354 < x < 355.58.

diluculo:
The approximation of `sinh^2(x)` is used if `x >= 22.0`. This value is taken from freeBSD.

cdrnet:
Great, thanks a lot once more!

PR #542, opened 2018-01-20
Argument checks

jnyrup:
`Jaccard` already throws if `a` or `b` is null, so no need to check them again.
And if the null initial null checks weren't there, `a.Length` or `b.Length` could throw an `NRE` before `(a == null && b == null)` was called

`Minimum` checked `initialGuess` twice instead of checking `initalPertubation`.

`AddVectorToScaledVector` checked `if (y.Length != x.Length)` twice.


cdrnet:
Good catches, thanks for the PR!

PR #543, opened 2018-01-20
Move variables into scopes

jnyrup:
The variables are only used inside a conditional scope, so I moved the variables inside those scopes.


cdrnet:
Appreciated, thanks!

PR #544, opened 2018-01-20
Use TryGetValue instead of ContainsKey+indexer

jnyrup:
This saves a call to `knownPivots` by using `TryGetValue` instead of using `ContainsKey` and the dictionary indexer.


cdrnet:
Thanks!

PR #545, opened 2018-01-20
Initialize list with known capacity

jnyrup:
This uses the capacity overload of the `List` constructor as we know that `orthogonalMatrix.ColumnCount` entries will be inserted.


cdrnet:
Thanks!

PR #555, opened 2018-03-13
Issue552 - Rearrange Cholesky Factor 

ggraham-412:
Rearranges Cholesky factor for less roundoff error when the input matrix has a large range of numbers.  Passes unit tests.


ggraham-412:
I'll revise and resubmit;  I thought I had removed the Issue552 project,

cdrnet:
Thanks!

PR #566, opened 2018-05-08
Added a new RandomSource: Xoshiro256StarStar.

colgreen:
This PRNG is the most recent evolution of Marsaglia's xorshift. It passes all known statistical tests, is fast to compute, has only 256 bits of state, and has good (and provable) distribution. 

[xoshiro / xoroshiro generators and the PRNG shootout](http://xoshiro.di.unimi.it/)
[Scrambled Linear Pseudorandom Number Generators](https://arxiv.org/abs/1805.01407)


cdrnet:
Great, thanks!

The code is nice to read, with just the right and useful comments - appreciated!

PR #577, opened 2018-06-19
NewtonRaphsonRn

mneukom:
Added a class NewtonRaphsonRn to find the zeroes in a system of equations.


eriove:
Just out of curiosity (I'm not a maintainer of this project), would it be better, and easier to discover, if this was an overload to the [existing](https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics/RootFinding/NewtonRaphson.cs) Newton-Raphson implementation? Since it is actually a completely different implementation (and hard to share code) the actual implementation could be delegated to an internal class that would look just like this one.

Just some thoughts...

mneukom:
Actually the Broyden class already does something similiar.. Closing this.

PR #584, opened 2018-07-10
AutoCorrelation Method

TobiasGlaubach:
This is a pull request for a fast fft based implementation of the autocorrelation function for an array x. 

It is similar to the `numpy.correlate` function and matlabs `autocorr` function.  It is defined such that the output is normalized to ACF_x(k=0)=1.0 and ACF_x(k>0) in [-1...1]. The value for ACF_x(k=0) is always dropped before ACF is returned since it will always be 1.0.

I tested the method against `numpy.correlate` with error magnitudes of <1e-16 for all deterministic signals, and errors <0.3% for the rnd testcase. Currently I don't know why the rnd testcase has a higher error than the deterministic cases, but <0.3% is still well enough I think for a function as fuzzy as the autocorrelation. 

I generated testdata in python like this:

```
import numpy as np
def acfFun(v):
    acf = np.correlate(v, v, 'full')
    acf /= acf[n-1]
    return acf[n:]
```

The testfiles I used for testing were generated with these signals:

```
import numpy as np
from scipy import signal

np.random.seed(10)
n = 256

t = np.linspace(0, 1, n, endpoint=False)
y_rnd = np.random.randn(n)
y_sin = np.sin(2 * np.pi * 5 * t)
y_sqr = signal.square(2 * np.pi * 5 * t)
y_pwm = signal.square(2 * np.pi * 30 * t, duty=(np.sin(2 * np.pi * t) + 1)/2)

```


cdrnet:
What was the rationale to skip the zero lag, even though it is always one? Is there an advantage in typical follow up processing? Or is this common practice in some field? I'm wondering because it makes it quite a bit harder to reason about the parameters, e.g. when specifying the interesting lags in an array.

TobiasGlaubach:
Hi,
I implemented it that way before, since the lag of k=0 has no information gain. Also this implementation makes plotting a little bit easier, if the autocorrelation is much smaller than 1.

It has no influence on the precision ( I simply skip the first element when I normalize and set the output). Back in the days it was just the more convenient approach for what I needed. I see your point that it makes the input to output lag definition a bit inconsistent. Also mathematically speaking not skipping the first element would be the more proper signature for the function.

I can change that and add a pull request for that. OK?

PR #588, opened 2018-07-17
Polynomial class

TobiasGlaubach:
This is my first implementation for a `Polynomial `class for numerical real valued polynomials as proposed [here](https://discuss.mathdotnet.com/t/math-net-polynomial-class-proposal/678).

The class and tests are partly influenced by pythons `numpy.polynomial.polynomial` functions. 

Until now, only real (`double`) valued polynomials are supported. Maybe extending the class similar to the `Matrix<T>` or `Vector<T>` classes would be a good next step.

The root finding algorithm is based on an approach similar to the algorithm discussed [here](https://www.ams.org/journals/mcom/1995-64-210/S0025-5718-1995-1262279-2/S0025-5718-1995-1262279-2.pdf)

Usage Example:
```
var x = new double[] {5, 4, 3, 0, 2};
var p = new Polynomial(x);

var s = p.ToString();   // "2 x^3 + 0 x^2 + 3 x^2 + 4 x^1 + 5"
var r = p.GetRoots();
/*
{System.Numerics.Complex[4]}
    [0]: {(0.701550016667397, 1.39817022748844)}
    [1]: {(0.701550016667397, -1.39817022748844)}
    [2]: {(-0.701550016667396, 0.727643365074517)}
    [3]: {(-0.701550016667396, -0.727643365074517)}
*/
```

This pull request is a base requirement for some LTI classes I want to implement in Future

Any hints, proposals or questions are very welcome.


cdrnet:
Thanks!

PR #590, opened 2018-08-03
Fixing Correlation.Auto Signatures (Issue #598)

TobiasGlaubach:
This pull request will fix the signatures of the Correlation.Auto methods.
Inputs now properly correspond to outputs, since the `ACF(k=0)` value is not hidden before returning the ACF.
Also it fixes a bug I found along the way in the `public static double[] Auto(IEnumerable<double> x, int[] k)` method.


cdrnet:
Thanks!

PR #599, opened 2018-10-13
A preliminary full implementation of the Black-Scholes model for European options...

gocchionero:
...including greek parameters and leverage computations.

I have read your guidelines which said not to worry too much about a "final" version, as you are probably going to refactor. So, I put the working code in one file with some inline documentation.


cdrnet:
Thanks!

PR #601, opened 2018-10-17
Airy, Bessel, and Hankel functions

diluculo:
Following Bessel-related functions are added:

- Airy function Ai
- Airy function Bi
- Bessel function of the first kind
- Bessel function of the second kind
- Modified Bessel function of the first kind
- Modified Bessel function of the second kind
- Spherical Bessel function of the first kind
- Spherical Bessel function of the second kind
- Hankel function of the first kind
- Hankel function of the second kind

I needed Bessel functions of real order and complex argument as a solution to specific diffusion equations, but I couldn't find any related C # code. So I decided to translate the [AMOS Fortran package](http://www.netlib.org/amos/) by hand. The codes may have mistakes, but I confirm that there is no problem within the scope of UnitTest at least.


cdrnet:
Thanks a lot, this is very welcome!

cdrnet:
I wonder, wouldn't it make sense to have separate functions instead of the expScaled parameter? Similar to `Gamma` vs `GammaLn` (not exactly the same, but a similar kind of variation).

cdrnet:
I agree that an `e` suffix is not that nice. What about something more intuitive and readable than just an `e`?

Or if we stick with a flag, what about using an enum instead of a boolean? In linear algebra we also have similar parametrizations e.g. to clarify the behavior on zeros (relevant e.g. in higher order functions over sparse vectors) - over the time we had both separate functions and boolean flags, and in the end ended up with enum flags (`Zeros` and `ExistingData`).

diluculo:
I agree the enum flag. Separated scaling functions can also be useful to someone. What about `Scaled` prefix?

```
public enum Scaled
{
    Unity = 0,
    Exponential = 1
}

BesselJ(double v, Complex z, Scaled scaled = Scaled.Unity)
ScaledBesselJ(double v, Complex z)
```

PR #604, opened 2018-11-07
Special Functions: cleanup Bessel-type functions

diluculo:
PR according to the discussion in #601. 

1. `Scale` enum was added and used instead of the `expScaled` boolean flag.
2. Scaled Bessel functions were added separately.
3. `HankelH1` and `HankelH2` return usually complex values, so their real argument functions were removed.
4. Scaling options were removed from `SphericalBesselJ` and `SphericalBesselY`. - It is hard to find libraries that supports scaled spherical Bessel functions.

`Scaled` prefix was used temporarily for exponentially scaled Bessel functions. Please suggest me a better suffix or prefix.

### DIFF
--- a/src/Numerics.Tests/SpecialFunctionsTests/BesselTests.cs
+++ b/src/Numerics.Tests/SpecialFunctionsTests/BesselTests.cs
@@ -251,7 +251,7 @@ namespace MathNet.Numerics.UnitTests.SpecialFunctionsTests
         public void BesselIRatioExact(int n, double zr, double zi, double cyr, double cyi, int decimalPlaces)
         {
             var z = new Complex(zr, zi);
-            var actual = SpecialFunctions.BesselI(n + 1, z, true) / SpecialFunctions.BesselI(n, z, true);
+            var actual = SpecialFunctions.BesselI(n + 1, z, SpecialFunctions.Scale.Exponential) / SpecialFunctions.BesselI(n, z, SpecialFunctions.Scale.Exponential);
             AssertHelpers.AlmostEqualRelative(new Complex(cyr, cyi), actual, decimalPlaces);
         }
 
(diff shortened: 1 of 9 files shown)


diluculo:
In the 3rd commit, Kelvin functions were added.

- `KelvinBe(nu, x)` - Kelvin function of the first kind
- `KelvinBer(nu, x)`, `KelvinBer(x)` - Kelvin function ber, the real part of the KelvinBe
- `KelvinBei(nu, x)`, `KelvinBei(x)` - Kelvin function bei, the imaginary part of the KelvinBe
- `KelvinBerPrime(nu, x)`, `KelvinBerPrime(x)` - derivative of the Kelvin function ber
- `KelvinBeiPrime(nu, x)`, `KelvinBeiPrime(x)` - derivative of the Kelvin function bei
- `KelvinKe(nu, x)` - Kelvin function of the second kind
- `KelvinKer(nu, x)`, `KelvinKer(x)` - Kelvin function ker, the real part of the KelvinKe
- `KelvinKei(nu, x)`, `KelvinKei(x)` - Kelvin function kei, the imaginary part of the KelvinKe
- `KelvinKerPrime(nu, x)`, `KelvinKerPrime(x)` - derivative of the Kelvin function ker
- `KelvinKeiPrime(nu, x)`, `KelvinKeiPrime(x)` - derivative of the Kelvin function kei

diluculo:
To close this, we have to determine how to handle the exponentially scaled Bessel functions. If we don't have a good prefix or suffix to the functions. IMHO, it would be better to use the `Scale` flag.

cdrnet:
I think I prefer separate functions, but with a Scaled-suffix instead of a prefix.

cdrnet:
I've already applied the change - I hope you're ok with it.

cdrnet:
Thanks again!

diluculo:
@cdrnet Great. Many thanks.

Issue #485, opened 2017-04-12
Improve combinatorics functionality in mathnet-numerics


s952163:
This is a feature request to improve combinatorics related functionality in MathNet.Numerics. The [combinatorics namespace](https://numerics.mathdotnet.com/api/MathNet.Numerics/Combinatorics.htm) does contain the functions to calculate the number of possible combinations/variations/permutations. It also contains the functions to generate a random sample of those. However unfortnately it lacks functionality to actually generate all the combinations etc. For example see this [SO post](http://stackoverflow.com/questions/42184242/f-9-names-into-disjointed-groups-of-2-3-and-4), so one ends up using recursion or list comprehensions. F# 4.1 has added `List.allPairs` for the Cartesian product.

As this is the go-to library for statistical and numerical analysis it would be great if Mathnet.Numerics could generate the actual variations/permutations/combinations with and without repetitions. In .NET currently this library does: https://github.com/eoincampbell/combinatorics . For examples see: https://www.codeproject.com/Articles/26050/Permutations-Combinations-and-Variations-using-C-G and [Combinatorics in .NET Part 1](http://trycatch.me/combinatorics-in-net-part-i-permutations-combinations-variations/).

gsomix:
I'll deal with it.

topinfrassi01:
Hi,

I'd be interested in tackling this issue. I just want to make sure about some things about the request.

So, for example, this method exists 

`public static IEnumerable<T> SelectCombinationWithRepetition<T>(this IEnumerable<T> data, int elementsToChoose, System.Random randomSource = null)`

It selects a random combination for the data. This very issue is about, in this example, generating **all** the possible combinations with repetitions in the enumerable?

PR #607, opened 2018-11-20
Add combinatorial generation algorithms

danielsword:
This PR is intended to satisfy some of the requests made in #485.

What's there:
* Methods for:
   - Generating all combinations for a given set and subset size, *without replacement*.
   - Generating all permutations for a given set, *without replacement*.
   - Inverting a permutation.
* Some tests.

Missing:
* Variations
* Repetitions

I'm happy to work on the missing pieces, but would appreciate any feedback first!

### DIFF
--- a/src/Numerics.Tests/CombinatoricsTests/PermutationTests.cs
+++ b/src/Numerics.Tests/CombinatoricsTests/PermutationTests.cs
@@ -39,6 +39,21 @@ namespace MathNet.Numerics.Tests.CombinatoricsTests
             Assert.AreEqual(expected,count);
         }
 
+        [TestCase(1,1)]
+        [TestCase(2,1)]
+        [TestCase(3,3)]
+        [TestCase(4,2)]
+        [TestCase(4,3)]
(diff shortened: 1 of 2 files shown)


danielsword:
I'm specifically (and very) interested in any feedback related to some of the comments in the code about lazy evaluation.

For example:

```C#
var permutations = Combinatorics.GenerateAllPermutations(n).ToList();
```

is just going to end disastrously for anybody who tries it.

eriove:
I've experienced the same problem with IEnumerable's that shouldn't be enumerated from end to end. One way of solving it would be to return a concrete type with an implementation of `ToList()` which throws an exception for larger numbers of permutations. That would at least prevent the most causal mistakes. That also gives you the opportunity to use [Debugger Display Attributes](https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/enhancing-debugging-with-the-debugger-display-attributes) to prevent the enumeration to crash Visual Studio when the watch window tries to enumerate the permutations.

PR #612, opened 2018-12-14
inline _factorialCache to a static readonly field

JeWaVe:
Assembly size should not be an issue in this code.
Should be a bit faster, since _factorialCache should now stay in the data segment (cached) instead of global memory
Values are store as hexadecimal constants, to save all bits, so new version returns the exact same values as before


cdrnet:
I agree. Thanks!

PR #614, opened 2018-12-30
Nonlinear least squares fitting

diluculo:
The Levenberg-Marquardt and the trust region dogleg algorithms have been implemented.

Notes:
- Initially, I started this PR with `IObjectiveFunction`, but I think it would be better to separate EvaluateAt() with EvaluateFunction() and EvaluateJacobian() to know the number of function/jacobian calls, so I added `IObjectiveModel`. The `Parameters` and `Residue` of `IObjectiveModel` are equivalent to the `Point` and `Value` of `IObjectiveFunction`, respectively. 
- To get more useful information, `ModelMinimizationResult` has been added instead of the `MinimizationResult`. 
- If necessary, they can be refactored by extending the originals.
- When a user-supplied analytical Jacobian is not given, the finite difference method is used to calculate the numerical Jacobian. When [AccuracyOrder](https://en.wikipedia.org/wiki/Finite_difference_coefficient) is even, the central finite difference method is applied. Otherwise, the forward difference method is used.
- To handle the box constrained (lower bound < parameter < upper bound) minimization, we need to transform the bounded parameters to unbounded. I found a very simple math in the [LMFIT](https://lmfit.github.io/lmfit-py/bounds.html)  (license: BSD-3). (*1)
- Some tests have been added.

### EDIT
(*1) The transformation rule was originally from [MINUIT](https://root.cern.ch/download/minuit.pdf).


diluculo:
Finished. I have modified this PR to make it as close as possible to the structure of the existing minimizers. Please review it. Sorry for so many post-commits.

cdrnet:
Thanks, and sorry for the delay!

PR #633, opened 2019-04-25
Issue 632, NelderMeadSimplex Does Not Always Return Best Evaluated Point

EricScottSalem:
Immediately prior to completion, re-evaluate objective function with best vertex to set proper state.


cdrnet:
Thanks!

PR #640, opened 2019-05-29
Implemented and tested Tukey window

MarcoRoss84:
Hi,

I am working for Philips and we intend to use MathNet.Numerics in one of our products and contribute some of the extensions we require back to the community. I'd like to start with this very simple extension to the available window functions: The Tukey window.

Regards, Marco Ross.


cdrnet:
Thanks, especially also for providing the tests!

The build error is unrelated to the change.

PR #654, opened 2019-08-08
Add Skewed Generalized T distribution and Skewed Generalized Error di…

MikaelUmaN:
- Adds Skewed Generalized T distribution as introduced by Theodossiou (1998)
- Adds Skewed Generalized Error distribution needed to cover cases where parameter q is not finite

This enables derived distributions such as:

1. Skewed Normal distribution
2. Skewed T distribution

The implementation is based on the package sgt in R, please see the [viginette](https://cran.r-project.org/web/packages/sgt/vignettes/sgt.pdf). Tests of density, cdf, inverse cdf have all been done compared to the R package using a number of parameterizations.

Note: In the R package, there are two options "mean.cent" and "var.adj". These affect the location and scale of the distribution. If they are true, then the location parameter will equal the mean and the scale parameter squared will equal the variance. **For simplicity, I have chosen to always set them equal to true**. 

The purpose is in general to allow for skewed probability distributions which are of praticular interest in some financial markets.

Possible issues:

- No stand-alone tests for Skewed Generalized Error Distribution. This is tested implicitly via the Skewed Generalized T distribution tests.
- No formulas for entropy.
- Not sure how to create an equivalence test with the existing StudentT distribution. I would have liked to create a test in the same manner as I have done for the Laplace distribution where a given parameterization of the Skewed Generalized T distribution can be shown to be equal to the existing StudentT distribution. This can be done in R by using another parameterization and setting mean.cent=False, var.adj = sqrt(2), but as stated above I always have them as True.

**Running in R**
If you want to check the R code. You can execute:

```r
install.packages("sgt")
library(sgt)
dsgt(1.2, mu=0, sigma=1, lambda=0, p=2, q=9)
```

Please aid with advice/help/feedback/comments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ out/
 .vs
 .vscode
 .idea
-.ionide/
+.ionide
 
 # Test & Analysis Results
 [Tt]est[Rr]esult*/
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ out/
 .vs
 .vscode
 .idea
+.ionide/
 
 # Test & Analysis Results
 [Tt]est[Rr]esult*/
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -5,6 +5,11 @@
     <!-- make MSBuild track this file for incremental builds. -->
     <!-- ref https://blogs.msdn.microsoft.com/msbuild/2005/09/26/how-to-ensure-changes-to-a-custom-target-file-prompt-a-rebuild/ -->
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+
+    <DetectedMSBuildVersion>$(MSBuildVersion)</DetectedMSBuildVersion>
+    <DetectedMSBuildVersion Condition="$(MSBuildVersion) == ''">15.0.0</DetectedMSBuildVersion>
+    <MSBuildSupportsHashing>false</MSBuildSupportsHashing>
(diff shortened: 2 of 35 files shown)


MikaelUmaN:
I am not sure why the Packages.Restore file is updated/changed. Is it because of a new version of Paket? Please advice if I have to do something different. I have seen in some other pull requests that that file causes conflicts.

cdrnet:
Sorry for the delay and the confusion around Paket and Travis. It seem the latter was an intermittent issue and has been resolved in the meantime.

It seems the sampling test `SkewedGeneralizedTTests.CanSample` currently fails. The AdjustScale function called by InvCDF cannot properly handle q=infinity, causing NaN. Any idea what is going wrong here?

MikaelUmaN:
Thanks.

I’ll have a look at it when I get some time and post back !

MikaelUmaN:
@cdrnet 

The problem was with the static InvCDF not just using a simpler distribution when the parameters were one of the extremes, e.g. q=Inf.

Should be fixed now.

MikaelUmaN:
@cdrnet

Sorry have you had a chance to look at Merging this yet?

Happy to adapt to latest master commit if I get some feedback that the pull request will be accepted.

cdrnet:
I finally found some time to look into this. I've rebased the PR and merged it into master (GitHub does not recognize it due to the rebasing). Thanks again for this contribution!

I had to do a small change on how the Sample functions are wired together, to be more consistent with the rest of the library, and to resolve a stack overflow.

MikaelUmaN:
Thanks a bunch:)!!

cdrnet:
Released in v4.11.0.

PR #655, opened 2019-09-11
Integration by substitution

diluculo:
Implemented the integration by substitution to integrate a function that has limits infinite. The function must be decay rapidly to 0 as x -> +/- infinity.


diluculo:
Finally, this PR has following functions:
- Integrate.GaussLegendre
- Integrate.GaussKronrod
- Integrate.DoubleExponential

and complex versions:
- ContourIntegrate.GaussLegendre
- ContourIntegrate.GaussKronrod
- ContourIntegrate.DoubleExponential

diluculo:
The Gauss-Kronrod quadrature pairs can be calculated very accurately by means of the Clenshaw algorithm. Finished. Please review to merge.

diluculo:
This PR may have 3 issues.

1.  Gauss-Kronrod quadrature rule 

There are 3 major methods to get abscissas and weights for Gauss-Kronrod quadrature rule.
   -	Patterson, Thomas NL. "The optimum addition of points to quadrature formulae." Mathematics of Computation 22.104 (1968): 847-856.
   -	Piessens, Robert, and Maria Branders. "A note on the optimal addition of abscissas to quadrature formulas of Gauss and Lobatto type." Mathematics of Computation (1974): 135-139.
   -	Laurie, Dirk. "Calculation of Gauss-Kronrod quadrature rules." Mathematics of Computation of the American Mathematical Society 66.219 (1997): 1133-1145.

I implemented the Patterson algorithm in this PR. It requires successive recursions, so the precision in coefficients of polynomials is truncated, and we fails to get abscissas and weights  in higher order > ~ 300. The Piessens algorithm is more stable but under GNU LGPL license.

2. Complex integrations

They support integral of a complex integrand to speed up the calculation. 

- `integral {f(x) + j g(x)} dx  = integral f(x) dx + j integral g(x) dx`

So, the `ContourIntegrate` is not right name. However, I hope someone will expand them to complex contour integration using waypoints someday.

3. Integration by substitution

To integrate a function on the range of (-oo, oo), (-oo, b], or [a, oo), I added some change of variable rules. But it is true these are not the best to all integrands. Therefore, we can hand these over to users.

src/Numerics/Integrate.cs
```diff
@@ -45,23 +46,11 @@ public static class Integrate
...
-        /// Approximation of the definite integral of an analytic smooth function on a closed interval.
-        /// </summary>
-        /// <param name="f">The analytic smooth function to integrate.</param>
-        /// <param name="intervalBegin">Where the interval starts, inclusive and finite.</param>
-        /// <param name="intervalEnd">Where the interval stops, inclusive and finite.</param>
-        /// <returns>Approximation of the finite integral in the given interval.</returns>
-        public static double OnClosedInterval(Func<double, double> f, double intervalBegin, double intervalEnd)
-        {
-            return DoubleExponentialTransformation.Integrate(f, intervalBegin, intervalEnd, 1e-8);
-        }
```
cdrnet: It seems removing this overload would be a breaking change (optional arguments are not actually optional once compiled). I think this should be reverted.

cdrnet:
Thanks a lot! I think the issues are ok.

I noticed one small but breaking change which we'd need to fix before I can bring this into master. See review above.

cdrnet:
Actually, I've fixed it on my side. Thanks.

PR #659, opened 2019-10-03
Detect early completion in BfgsBMinimizer

### LINKED ISSUE #658: BfgsBMinimizer fails to converge on simple boundary solution
We've seen this optimisation method fail to converge on many examples, including the very simple example I've created a failing test for here: https://github.com/tmct/mathnet-numerics/commit/1f8d0536d388e47e2b8f9ccf91f21e6870ea0f5a (second test fails to converge)

This just tries to minimise `x^2 + y^2` over [1,2] x [1,2], and it feels like it should arrive at [1,1] fairly quickly from that...

I may try to fix this and create a PR.

tmct:
Resolves #658


tmct:
Closing and re-opening because I don't trust the Travis failure

cdrnet:
Thanks!

PR #662, opened 2019-10-14
Fractional calculus

diluculo:
The Riemann–Liouville fractional derivative is added.

References:
- [Differintegral](https://en.wikipedia.org/wiki/Differintegral)
- [Fractional Calculus](https://en.wikipedia.org/wiki/Fractional_calculus)


cdrnet:
Finally merged - thanks a lot!

Issue #679, opened 2019-12-18
Hypergeometric.CDF can return values greater than 1.0 (MathNet.Numerics v4.9.0)


manor:
```c#
using System;
using MathNet.Numerics.Distributions;

namespace BugReport
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine(Hypergeometric.CDF(1953, 453, 84, 47));  // 0.9999999999948128
            Console.WriteLine(Hypergeometric.CDF(1953, 453, 84, 48));  // 1.0000000000004392
            if(Hypergeometric.CDF(1953, 453, 84, 48) > 1.0)
            {
                Console.WriteLine("CDFs should never yield values greater than 1.0!!!");
            }
        }
    }
}
```

cdrnet:
I've merged in the workaround - thanks for the analysis and the PR!

PR #680, opened 2019-12-19
Ensure Hypergeometric.CDF <= 1.0 (Issue #679)

manor:
This is, to be fair, a kludge -- it would be more efficient to ensure that such a situation can never arise in the first place. Proposed as a temporary(?) solution to issue #679


cdrnet:
Thanks!

PR #681, opened 2020-01-19
Tests for Moore-Penrose Pseudo-Inverse and bugfix for Pseudo-Inverse of complex matrices

TillAlex:
Moore-Penrose Pseudo-Inverse currently gives wrong results for complex matrices. This PR adds tests for Pseudo-Inverses and fixes the implementation for complex matrices.

The tests fails for DiagonalMatrix because there seems to be a bug in the multiplication of non square diagonal matrices with sparse matrices. Unfortunately I am not able to fix this bug.


TillAlex:
Accidently added another commit. Here is the correct PR: #682

PR #682, opened 2020-01-20
Tests for Moore-Penrose Pseudo-Inverse and bugfix for Pseudo-Inverse of complex matrices

TillAlex:
Moore-Penrose Pseudo-Inverse currently gives wrong results for complex matrices. This PR adds tests for Pseudo-Inverses and fixes the implementation for complex matrices.

The tests fails for DiagonalMatrix because there seems to be a bug in the multiplication of non square diagonal matrices with sparse matrices. Unfortunately I am not able to fix this bug.

### DIFF
--- a/src/Numerics.Tests/LinearAlgebraTests/Complex/MatrixTests.Arithmetic.cs
+++ b/src/Numerics.Tests/LinearAlgebraTests/Complex/MatrixTests.Arithmetic.cs
@@ -820,6 +820,20 @@ namespace MathNet.Numerics.UnitTests.LinearAlgebraTests.Complex
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


cdrnet:
Thanks!

The tests seem to hit an unrelated bug (diagonal matrice).

cdrnet:
I finally found some time to look into this. I've rebased the PR and merged it into master (GitHub does not recognize it due to the rebasing). The tests also revealed another issue on non-square diagonal sparse mixed products. Thanks again for this contribution!

cdrnet:
Released in v4.11.0

TillAlex:
Thanks!

Issue #684, opened 2020-01-25
Incorrect weights in NonlinearObjectiveFunction


keithyipkw:
Should  `jacobianValue[i, j] * L[j]` being using `L[i]`?

NonlinearObjectiveFunction gives poor fittings when weights are set. The gradients and Hessians are wrong.

```
            L = (weights == null)
                ? null
                : Weights.Diagonal().PointwiseSqrt();
```

```
            // weighted jacobian
            for (int i = 0; i < NumberOfObservations; i++)
            {
                for (int j = 0; j < NumberOfParameters; j++)
                {
                    if (IsFixed != null && IsFixed[j])
                    {
                        // if j-th parameter is fixed, set J[i, j] = 0
                        jacobianValue[i, j] = 0.0;
                    }
                    else
                    {
                        jacobianValue[i, j] = (Weights == null)
                            ? jacobianValue[i, j]
                            : jacobianValue[i, j] * L[j];
                    }
                }
            }
```

diluculo:
You are right.

cdrnet:
Thanks for the analysis and the fix!

PR #685, opened 2020-01-27
Fix incorrect weighted Jacobian in NonlinearObjectiveFunction

diluculo:
PO for #684


cdrnet:
Great, thanks for the fix!

PR #688, opened 2020-02-19
Fixed incorrectly implemented row reducing

kayanme:


cdrnet:
Good catch - thanks for providing the fix!

PR #692, opened 2020-03-24
fix the error on index of observations' weights

zhaochenxi:
"L" is the weights for the observations.
"i" in EvaluateJacobian() method is the NumberOfObservations. so the "L"'s index should be "i".


diluculo:
Duplicate of #685

cdrnet:
I've already merged in the other fix - but thanks a lot for the analysis and for providing the PR!

Issue #690, opened 2020-03-11
Failures of MathNet.Numerics.Tests.MKL


diluculo:
I have tested `MathNet.Numerics.Tests.MKL` project with `MathNet.Numerics.MKL.Win` (v2.3.0, MKL 2018.0 R1). 

```
Math.NET Numerics Configuration:
Version 4.9.0
Built for .Net Framework 4.6.1
Linear Algebra Provider: Intel MKL (x64; revision 12; MKL 2018.0 Update 1)
Fourier Transform Provider: Intel MKL (x64; revision 12; MKL 2018.0 Update 1)
Max Degree of Parallelism: 8
Parallelize Elements: 300
Parallelize Order: 64
Check Distribution Parameters: True
Thread-Safe RNGs: True
Operating System: Microsoft Windows NT 6.2.9200.0
Framework: 4.0.30319.42000
```

There are 43 failed tests and 2 errors. Most of them may be minor and related to accuracy : 

```
1) Failed : MathNet.Numerics.UnitTests.DistributionTests.Multivariate.MatrixNormalTests.ValidateDensity
Not equal within 16 places. Expected:0.000156829273664912; Actual:0.000156829273664912
   위치: MathNet.Numerics.UnitTests.AssertHelpers.AlmostEqualRelative(Double expected, Double actual, Int32 decimalPlaces)

7) Failed : MathNet.Numerics.UnitTests.LinearAlgebraTests.Complex.Factorization.UserCholeskyTests.CanFactorizeRandomMatrix(1)
  Expected: (0.353088257803903, 0)
  But was:  (0.353088257803903, -8.01291052930066E-21)
   위치: MathNet.Numerics.UnitTests.LinearAlgebraTests.Complex.Factorization.UserCholeskyTests.CanFactorizeRandomMatrix(Int32 order)
```
And some are related with `LUFactor()` ( => [`LAPACKE_dgetrf`](https://software.intel.com/en-us/mkl-developer-reference-c-getrf) function of Intel MKL):

```
40) Error : MathNet.Numerics.UnitTests.RootFindingTests.BroydenTest.Seveneq3b
MathNet.Numerics.InvalidParameterException : An invalid parameter was passed to a native method, parameter number : 4
   위치: MathNet.Numerics.Providers.LinearAlgebra.Mkl.MklLinearAlgebraProvider.LUFactor(Double[] data, Int32 order, Int32[] ipiv)
[...]

PR #695, opened 2020-04-09
Intel MKL direct sparse solver for finite element method problems

diluculo:
- Added MKL direct sparse solver for solving sparse systems of linear equations.
   - `SparseSolverControl.Provider.Solve(...)`
- Added methods to build a matrix from a COO, CSR, or CSC format.
   - `Matrix<T>.Build.SparseFromCoordinateFormat(...)`
   - `Matrix<T>.Build.SparseFromCompressedSparseRowFormat(...)`
   - `Matrix<T>.Build.SparseFromCompressedSparseColumnFormat(...)`
- Added related unit tests.

Possible issues:
- In capabilities.cpp of MKL project, I didn't know the meaning of the versions, so I selected versions for direct sparse solver randomly.
- `ManagedSparseSolverProvider` was not implemented. 
- `Providers\Experimental.cs` was for my testing purpose. Actually I couldn't decide where to add it. Please remove it before merge.

Any comments on the code and API would be greatly appreciated.

----------------
FYI, my steps to build MathNet.Numerics.MKL.sln are:
1. I have no Intel Parallel Studio, so I first downloaded to install the [Intel MKL library](https://software.intel.com/en-us/mkl).    
   - Registration is needed. 
   - I choosed the Intel Math Kernel Library 2019 Update 5 version.
2. Replace all `INTEL_DEV_REDIST` with `MKLProductDir` in the MKLWrapper.vcxproj. 
3. Open Command Prompt and go to the MathNet.Numerics repository
4. Run `build.cmd MklWinBuild` or `build.cmd MklWinAll`.
5. Find to keep `\out\MKL` folder.
6. To run unit test, open Mathnet.Numerics.MKL.sln.
7. Build Numerics, TestData and Numerics.Tests.MKL projects. Note it will delete the `\out\MKL` folder.
8. Copy the `\x86` and `x64` folders from the folder archived in step 5  to `\src\Numerics.Tests\bin\Debug\net461` folder.
9. Run the unit tests. It will show lots of failures(#690), but the MKL package works for sure.


cdrnet:
This looks great, thanks a lot!

Would it be possible to rebase this on master? Otherwise I can also do it on my side.

diluculo:
@cdrnet, rebased.

- direct solver codes were removed from Benchmark.
- `Providers\Experimental.cs` was removed.

By the way, I think the enums, defined in `ISparseSolverProvider.cs`, can have better names. All refactoring is welcome.
- `Symmetricity` instead of `DssMatrixStructure`
- `Definiteness` instead of `DssMatrixType`
- `Transpose` instead of `DssSystemType` 
- And if details of `DssStatus` are not required, we can just throw exception instead.

PR #697, opened 2020-04-26
Fix Issue 642: Throw exception for invalid accuracy values.

RyanGrange:
Throw exception for invalid accuracy values.
Document value requirements.


cdrnet:
Thanks!

PR #706, opened 2020-06-23
Beta-Binomial distribution, and related functions

ajwillshire:
Adding the Beta-Binomial distribution. [https://en.wikipedia.org/wiki/Generalized_hypergeometric_function](https://en.wikipedia.org/wiki/Beta-binomial_distribution)

Also added the Generalized Hypergeometric Function as I thought it was required for calculating the CDF. However, most implementations just traverse over the PMF and sum. (Plus, I don't think Wikipedia is correct in the stated formula for the CDF). [https://en.wikipedia.org/wiki/Generalized_hypergeometric_function](https://en.wikipedia.org/wiki/Generalized_hypergeometric_function)

This also required adding the Rising and Falling Factorials [https://en.wikipedia.org/wiki/Falling_and_rising_factorials](https://en.wikipedia.org/wiki/Falling_and_rising_factorials)

### DIFF
--- a/src/Numerics/Distributions/BetaBinomial.cs
+++ b/src/Numerics/Distributions/BetaBinomial.cs
@@ -1,0 +1,416 @@
+﻿// <copyright file="BetaBinomial.cs" company="Math.NET">
+// Math.NET Numerics, part of the Math.NET Project
+// http://numerics.mathdotnet.com
+// http://github.com/mathnet/mathnet-numerics
+//
+// Copyright (c) 2009-2014 Math.NET
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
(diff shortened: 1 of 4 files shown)


src/Numerics/Distributions/BetaBinomial.cs
```diff
@@ -0,0 +1,416 @@
...
+
+        public System.Random RandomSource
+        {
+            get => _random;
+            set => _random = value ?? SystemRandomSource.Default;
+        }
+        /// <summary>
+        /// Gets the mean of the distribution.
+        /// </summary>
+        double IUnivariateDistribution.Mean => (_n * _a) / (_a + _b);
```
cdrnet, line 144: What is the reasoning for choosing explicit interface implementations here, i.e. for making them unavailable on the type?

cdrnet:
Looks great, thanks!

ajwillshire:
No probs.

Re the question of explicit interface implementations, I just tried to copy the existing style of (I think) the Negative Binomial distribution. I'm a bit out of my comfort zone, so happy for changes to be made as required.

Thanks!

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

cdrnet:
No worries, I can go over them - the explicit interface implementation actually makes sense for those throwing exceptions. Unfortunately we cannot do that on the other existing ones as this would be a breaking change.

PR #707, opened 2020-06-29
Allow empty Vector and Matrix

### LINKED ISSUE #134: Support for 0 lenght Vector and 0 by 0 Matrix
Not having 0 length Vectors and 0 by 0 Matrices creates many issues and porting issues from other libraries / code that handled 0 length Lists/ Vectors.

I am not sure what the original reasons were for not supporting 0 length Vectors and Matrices.
Every time on creates a vector from an array or enumerable on needs to check if that is not an empty collection and if it is an empty collection one has to return  null instead of a 0 length Vector and this creates lots of nulls objects in the code.

Evangelink:
Fix #134 

This is an attempt at allowing empty (zero-length) `Vector` and `Matrix`. I am relying on the existing tests to find any issue with what I have done as I am not enough of an expert to see if there are any unwanted consequences.

### DIFF
--- a/src/Numerics.Tests/LinearAlgebraTests/Complex/VectorTests.cs
+++ b/src/Numerics.Tests/LinearAlgebraTests/Complex/VectorTests.cs
@@ -63,7 +63,7 @@ namespace MathNet.Numerics.UnitTests.LinearAlgebraTests.Complex
             Assert.AreEqual(vector.Count, clone.Count);
             CollectionAssert.AreEqual(vector, clone);
         }
-        
+
         /// <summary>
         /// Can clone a vector using <c>IClonable</c> interface method.
         /// </summary>
@@ -198,7 +198,6 @@ namespace MathNet.Numerics.UnitTests.LinearAlgebraTests.Complex
         public void SizeIsNotPositiveThrowsArgumentOutOfRangeException()
         {
             Assert.That(() => CreateVector(-1), Throws.TypeOf<ArgumentOutOfRangeException>());
-            Assert.That(() => CreateVector(0), Throws.TypeOf<ArgumentOutOfRangeException>());
         }
 
         /// <summary>
--- a/src/Numerics.Tests/LinearAlgebraTests/MatrixTests.cs
+++ b/src/Numerics.Tests/LinearAlgebraTests/MatrixTests.cs
@@ -1,0 +1,77 @@
+﻿// <copyright file="VectorStorageCombinatorsTests.cs" company="Math.NET">
+// Math.NET Numerics, part of the Math.NET Project
+// http://numerics.mathdotnet.com
+// http://github.com/mathnet/mathnet-numerics
+//
+// Copyright (c) 2009-2016 Math.NET
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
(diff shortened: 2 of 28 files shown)


src/Numerics.Tests/LinearAlgebraTests/Complex/VectorTests.cs
```diff
@@ -63,7 +63,7 @@ public void CanCloneVector()
             Assert.AreEqual(vector.Count, clone.Count);
             CollectionAssert.AreEqual(vector, clone);
         }
-        
+
```
Evangelink, line 66: I have a VS extension removing the trailing whitespaces. Let me know if you want me to revert these changes.

src/Numerics.Tests/LinearAlgebraTests/MatrixTests.cs
```diff
@@ -0,0 +1,77 @@
...
+using MathNet.Numerics.LinearAlgebra.Storage;
+using NUnit.Framework;
+
+namespace MathNet.Numerics.UnitTests.LinearAlgebraTests
+{
+    [TestFixture, Category("LA")]
+    public class MatrixTests
+    {
+        [Test]
+        public void DenseMatrixBuilderMethos_ZeroLength_DoNotThrowException()
```
Evangelink, line 41: Let me know your preferences for grouping tests together.

src/Numerics.Tests/LinearAlgebraTests/MatrixTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            Assert.DoesNotThrow(() => Matrix<double>.Build.Dense(0, 0));
```
Evangelink, line 43: I have only tested with double as the base class is generic but I could add tests with the various supported types if you want to.

src/Numerics/LinearAlgebra/Storage/DenseVectorStorage.cs
```diff
@@ -115,9 +115,9 @@ public static DenseVectorStorage<T> OfVector(VectorStorage<T> vector)
 
         public static DenseVectorStorage<T> OfValue(int length, T value)
         {
-            if (length < 1)
+            if (length < 0)
             {
-                throw new ArgumentOutOfRangeException(nameof(length), string.Format(Resources.ArgumentLessThanOne, length));
+                throw new ArgumentOutOfRangeException(nameof(length), Resources.ArgumentNotNegative);
```
Evangelink, line 120: You have 2 messages that are relatively close: `ArgumentNotNegative` and `ArgumentMustBePositive`. The first one explicitly mention that zero is allowed while the second doesn't say anything. When reading the message about the "must be positive" case, I wouldn't have guessed this wasn't allowing zero.

Anyways, which message do you want to use? Shall I remove the other?

Evangelink:
That's weird, I have just noticed that some tests do not want to run locally. I'll have a look at what's going-on and fix them.

src/Numerics.Tests/LinearAlgebraTests/Complex/VectorTests.cs
cdrnet: No, that's fine - I have similar extension installed (but apparently not touched that file since)

src/Numerics/LinearAlgebra/Storage/DenseVectorStorage.cs
cdrnet: Semantically for integers I understand positive to not include zero, and non-negative to include zero. I noticed that this is explicitly mentioned in the text of `ArgumentPositive` but not in `ArgumentMustBePositive`.

Yes, unfortunately these resources are a bit of a mess currently, I need to clean them up. Or even get rid of them actually, since we don't have any translations anyway...

Evangelink: Chatting with some native English speaker friends, it seems that for Maths in English zero is neither included in positives nor negatives. So saying "must be positive" should exclude zero but saying "should not be negative" does include zero.

PR #711, opened 2020-06-30
Fix thread culture for some tests

### LINKED ISSUE #710: Some unit tests are failing on non-English OSes
Most of the `ValidateToString` and also some other tests are failing on non-English OSes as the expected result uses the english/invariant culture.

Evangelink:
Refactor code to force `CultureInfo.InvariantCulture` on the main thread for some tests.

Fix #710

### DIFF
--- a/src/Numerics.Tests/DistributionTests/Continuous/BetaScaledTests.cs
+++ b/src/Numerics.Tests/DistributionTests/Continuous/BetaScaledTests.cs
@@ -85,6 +85,7 @@ namespace MathNet.Numerics.UnitTests.DistributionTests.Continuous
         [Test]
         public void ValidateToString()
         {
+            System.Threading.Thread.CurrentThread.CurrentCulture = System.Globalization.CultureInfo.InvariantCulture;
             var n = new BetaScaled(1d, 2d, 0.0, 1.0);
             Assert.AreEqual("BetaScaled(α = 1, β = 2, μ = 0, σ = 1)", n.ToString());
         }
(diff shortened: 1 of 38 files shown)


Evangelink:
It'd be nice to have a new build axis using a different culture to ensure we don't introduce regression or new culture error in tests.

src/Numerics.Tests/DistributionTests/Continuous/BetaScaledTests.cs
```diff
@@ -85,6 +85,7 @@ public void BetaScaledCreateFailsWithBadParameters()
         [Test]
         public void ValidateToString()
         {
+            System.Threading.Thread.CurrentThread.CurrentCulture = System.Globalization.CultureInfo.InvariantCulture;
             var n = new BetaScaled(1d, 2d, 0.0, 1.0);
             Assert.AreEqual("BetaScaled(α = 1, β = 2, μ = 0, σ = 1)", n.ToString());
```
cdrnet, line 90: Just curious, are there really cultures where floating point numbers with 0 decimal part and a single digit only are rendered differently than the expected string?

cdrnet:
Thanks a lot for looking into this!

src/Numerics.Tests/DistributionTests/Continuous/BetaScaledTests.cs
Evangelink: I have fixed the tests failing on my machine but looking at it, that does seem weird. Let me try to run the test again.

Evangelink: Looks like I "fixed" one un-needed test or I had an unrelated issue with this test and I applied a wrong fix. Anyways I'll undo this one! Thanks for the catch!

cdrnet:
> It'd be nice to have a new build axis using a different culture to ensure we don't introduce regression or new culture error in tests.

Yes, I agree. Also we may need an overload for all these functions accepting an `IFormatProvider`.

PR #712, opened 2020-06-30
Fix RayleightTests.ValidateMaximum test to use testcase parameter

### LINKED ISSUE #708: Unused parameter in ValidateMaximum on RayleighTests.cs
The parameter `scale` is unused in method `ValidateMaximum` in `RayleigthTests.cs`: https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics.Tests/DistributionTests/Continuous/RayleighTests.cs#L197-L205

Evangelink:
Fix `RayleightTests.ValidateMaximum` test to use the scale parameter instead of the hardcoded value.

Fix #708


cdrnet:
Thanks!

PR #713, opened 2020-06-30
Fix GeometricTests.ValidateProbabilityLn to use expected value parameter

### LINKED ISSUE #709: Unused parameter in test ValidateProbabilityLn
The parameter `pln` is unused in test `ValidateProbabilityLn` in `GeometricTests.cs`: https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics.Tests/DistributionTests/Discrete/GeometricTests.cs#L177-L200

I am not sure if it should replace the left parameter on line 195 or line 199.

Evangelink:
Fix #709

### DIFF
--- a/src/Numerics.Tests/DistributionTests/Discrete/GeometricTests.cs
+++ b/src/Numerics.Tests/DistributionTests/Discrete/GeometricTests.cs
@@ -175,28 +175,21 @@ namespace MathNet.Numerics.UnitTests.DistributionTests.Discrete
         /// <param name="x">Input X value.</param>
         /// <param name="pln">Expected value.</param>
         [TestCase(0.0, -1, Double.NegativeInfinity)]
-        [TestCase(0.0, 0, 0.0)]
+        [TestCase(0.0, 0, Double.NegativeInfinity)]
         [TestCase(0.0, 1, Double.NegativeInfinity)]
         [TestCase(0.0, 2, Double.NegativeInfinity)]
         [TestCase(0.3, -1, Double.NegativeInfinity)]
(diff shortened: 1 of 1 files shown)


src/Numerics.Tests/DistributionTests/Discrete/GeometricTests.cs
```diff
@@ -175,28 +175,21 @@ public void ValidateProbability(double p, int x)
...
+        [TestCase(0.3, 0, Double.NegativeInfinity)]
         [TestCase(0.3, 1, -1.2039728043259360296301803719337238685164245381839102)]
-        [TestCase(0.3, 2, Double.NegativeInfinity)]
+        [TestCase(0.3, 2, -1.5606477482646686)]
         [TestCase(1.0, -1, Double.NegativeInfinity)]
         [TestCase(1.0, 0, Double.NegativeInfinity)]
-        [TestCase(1.0, 1, 0.0)]
+        [TestCase(1.0, 1, Double.NaN)]
         [TestCase(1.0, 2, Double.NegativeInfinity)]
         public void ValidateProbabilityLn(double p, int x, double pln)
```
Evangelink, line 189: It might be better to have 2 set of tests one using the expected result and the other doing the calculation.

cdrnet: It seems the one doing the computation was using exactly the same formula as the implementation, which was kind of pointless. So your changes toward using the provided expected value looks more useful to me.

cdrnet:
Thanks!

PR #724, opened 2020-08-10
Issue#721 - not-sorted and duplicate entries of COO format

diluculo:
PR for #721

### DIFF
--- a/src/Numerics/LinearAlgebra/Builder.cs
+++ b/src/Numerics/LinearAlgebra/Builder.cs
@@ -48,8 +48,12 @@ namespace MathNet.Numerics.LinearAlgebra.Double
             return new DenseMatrix(storage);
         }
 
-        public override Matrix<double> Sparse(SparseCompressedRowMatrixStorage<double> storage)
+        public override Matrix<double> Sparse(SparseCompressedRowMatrixStorage<double> storage, bool cleanup = false)
         {
+            if (cleanup)
+            {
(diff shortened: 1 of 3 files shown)


diluculo:
2nd commit:

- More comments are added.
- Explicit zeros are not intentionally removed.

src/Numerics/LinearAlgebra/Builder.cs
```diff
@@ -48,8 +48,12 @@ public override Matrix<double> Dense(DenseColumnMajorMatrixStorage<double> stora
             return new DenseMatrix(storage);
         }
 
-        public override Matrix<double> Sparse(SparseCompressedRowMatrixStorage<double> storage)
+        public override Matrix<double> Sparse(SparseCompressedRowMatrixStorage<double> storage, bool cleanup = false)
```
cdrnet, line 51: Unfortunately adding an optional parameter is a breaking change, so I'll have to rework this slightly.

diluculo: Please rework the line. Thank you.

cdrnet: I'll do - likely will have to become an overload.

cdrnet: Ah, I did not realize the affected members are actually internal (since the class is internal). All is fine after all. Thanks again!

PR #489, opened 2017-04-27
Unified optimization

eriove:
As I understand [this list](https://github.com/mathnet/mathnet-numerics/issues/283#issuecomment-136821287) in #283 there are no blocking issues to merge this into the main branch.

I did a rebase of the unified_optimization branch and fixed the merge problems. All tests pass. By merging this it will be easier to make new contributions for those who are interested in optimization and it will be usable for those that use the NuGet packages.

PR #728, opened 2020-08-27
Brent Minimizer

rfellers:
Hi all,

I recently was looking for a Brent minimizer when porting some python code to C# and I was hoping to find it here. `FindMinimum.OfScalarFunctionConstrained()` was pretty close, but given it is Golden Section under the covers, there were discrepancies with some inputs. I ended up porting the version from scipy [_minimize_scalar_bounded()](https://github.com/scipy/scipy/blob/v1.5.2/scipy/optimize/optimize.py#L1883).

I'd love to give back and contribute this port (and perhaps expand on it a little bit). Is this something of interest to the project? I've started a PR with a simple mock up of the rough API and tests I was thinking about (copied from GoldenSectionMinimizer).

Also, I noticed that there was quite a history with optimizers (#489, #173), so I thought I'd check first. I see that a BrentMinimizer class was considered in the past [optimization-1](https://github.com/mathnet/mathnet-numerics/blob/optimization-1/src/Numerics/Optimization/BrentMinimizer.cs). But it doesn't look like this has been updated in quite some time. Any insight here would be great.

### DIFF
--- a/src/Numerics.Tests/OptimizationTests/BrentMinimizerTests.cs
+++ b/src/Numerics.Tests/OptimizationTests/BrentMinimizerTests.cs
@@ -1,0 +1,60 @@
+﻿// <copyright file="BrentMinimizerTests.cs" company="Math.NET">
+// Math.NET Numerics, part of the Math.NET Project
+// http://numerics.mathdotnet.com
+// http://github.com/mathnet/mathnet-numerics
+//
+// Copyright (c) 2009-2017 Math.NET
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
+//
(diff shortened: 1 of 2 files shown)


eriove:
I was one of the people who worked on getting the existing minimizers/solvers into MathNet Numerics. I underestimated the effort left where I took it up and while we got some solvers merged there's a lot of things that could be improved and added.

To me this looks like a great initiative (I'm not a maintainer though).

src/Numerics.Tests/OptimizationTests/BrentMinimizerTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        [Test]
+        public void Test_ExpansionWorks()
+        {
+            var algorithm = new BrentMinimizer(1e-5, 1000);
+            var f1 = new Func<double, double>(x => (x - 3) * (x - 3));
+            var obj = ObjectiveFunction.ScalarValue(f1);
+            var r1 = algorithm.FindMinimum(obj, -2, 2);
+
+            Assert.That(Math.Abs(r1.MinimizingPoint - 3.0), Is.LessThan(1e-4));
```
blhough, line 57: The upper bound is x≤2 so it should not reach the global min at x=3.
I believe the expected `MinimizingPoint` is x=2

PR #729, opened 2020-08-27
Add extension methods that allow round above zero

KaptenJon:
Extension method to allow easy rounding of numerics


KaptenJon:
hope you like those extension methods. I do!

cdrnet:
Finally found some time to look at this. Yes, I can see how this can be useful - thanks!

PR #737, opened 2020-10-12
Updated golden section minimizer to avoid precision loss

dafuchs:
Recompute the middle point for each iteration to avoid loss of precision.


eHildy:
Just curious if there's been any movement on this?

cdrnet:
Looks good to me, thanks a lot!

Issue #727, opened 2020-08-24
TypeLoadException using MathNet.Numerics.Data.Matlab


Dblm0:
Given netCore3.1 console application with MathNet.Numerics.Data.Matlab (4.0.0) and MathNet.Numerics.Data.Text (4.0.0) usages, once I reference a netstandard lib which is using MathNet.Numerics (4.12.0), I got System.TypeLoadException on Matlab read or write call with message:
```
Could not load type 'MathNet.Numerics.Properties.Resources' from assembly 'MathNet.Numerics, Version=4.12.0.0, Culture=neutral, PublicKeyToken=null'.
```
Here is simple example to reproduce https://github.com/Dblm0/Math.NET-Test

There are no errors with MathNet.Numerics (4.11.0)

phihub:
We are having the same problem. 
I guess it was introduced by commit: 
[https://github.com/mathnet/mathnet-numerics/commit/43297a03668b2d14f94d35347eedeb421f9807b8]

The resources used in Data should be replaced by strings.

armwal:
Same here, MathNet.Numerics.Data.Matlab is unfortunately currently broken because of this. It would be great if the fix proposed above could be integrated!

cdrnet:
Ah, I completely missed this dependency. Thanks for reporting, and for the PR!

cdrnet:
Should be fixed in the 4.1 versions of the .Text and .Matlab packages, which now require Numerics 4.12.0 or newer.

PR #743, opened 2020-11-12
Removed Dependency to Mathnet.Numerics.Properties Resources

phihub:
I inlined the strings that were in the resources of Mathnet.Numerics in the Versions lower than 4.12.

This should fix Issue https://github.com/mathnet/mathnet-numerics/issues/727

This is my first PR, so please be gentle.


cdrnet:
Thanks a lot for looking into this and providing a fix!

PR #751, opened 2021-01-13
Implements IntroSort

bclehmann:
In case you're unaware IntroSort (aka Introspective Sort) is a hybrid sorting algorithm that is always at least as fast as QuickSort. It is the algorithm used by `Array.Sort` and several implementations of the C++ Standard Library.

The algorithm works like this:
- If the partition size is < 16 use insertion sort
- If the recursion depth is > 2 * log2(n) use heapsort
- Otherwise use quicksort (in this case a randomized quicksort)

I have not touched `Sorting.SortAll` because I think the most elegant way to write that function is to sort on the secondary comparer, followed by the primary comparer (similar to RadixSort). However, this requires a stable sorting algorithm, which IntroSort is not (by virtue of using QuickSort and HeapSort, which are both unstable). Choosing a stable sort like MergeSort would mildly degrade performance, but it would also allow a user to pass in a long list of comparers without requiring a long list of overloads.

### DIFF
--- a/src/Numerics.Tests/SortingTests.cs
+++ b/src/Numerics.Tests/SortingTests.cs
@@ -175,7 +175,8 @@ namespace MathNet.Numerics.UnitTests
             }
 
             // default sorting (Ascending)
-            list.Sort();
+            //list.Sort();
+            Sorting.Sort(list);
 
             // just check that the order is as expected, not that the items are correct
             for (var i = 1; i < list.Count; i++)
--- a/src/Numerics/Sorting.cs
+++ b/src/Numerics/Sorting.cs
@@ -474,16 +474,16 @@ namespace MathNet.Numerics
         /// <param name="recursions">Tracks the number of recursions entered.</param>
         static void IntroSort<TKey, TItem>(IList<TKey> keys, IList<TItem> satelliteData, IComparer<TKey> comparer, int left, int right, int recursions = 0)
         {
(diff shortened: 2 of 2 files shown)


Benny121221:
I don't know why it didn't work, but it looks like it could be the pipeline:

image

Benny121221:
[This overload](https://github.com/mathnet/mathnet-numerics/blob/a74ee1874b3b12a655b49bc9adfa201d0a9c21f7/src/Numerics/Sorting.cs#L143) (and some like it) seem have an unclear behaviour. The original implementation passed tests where `count` was actually a right bound, i.e. `index + count`. I think I'm misunderstanding what this overload is intended to do, as some linear algebra tests fail when introsort is used (with either `count` or `index + count`).

I reverted it to use the QuickSort implementation so as to pass tests until I see the intention with more clarity.

Corniel:
Shouldn't a benchmark prove that this faster than the current implementation?

bclehmann:
> Shouldn't a benchmark prove that this faster than the current implementation?

Probably. But since it's been a year and a half without activity, and I've completely forgotten about this PR, I'm not very interested in putting more work into a PR that might well be in vain.

There are problems with this PR that I'm not likely to end up fixing, so I think the best thing to do is probably close this, if someone later wants to make a similar change they can start from here if they want. If not, well this isn't exactly huge or too difficult, so even starting from scratch isn't a big loss.

PR #756, opened 2021-02-24
Logistic distribution

bobbyingram:
Adding the Logistic distribution: https://en.wikipedia.org/wiki/Logistic_distribution

I've tried to follow the pattern of the `Normal` distribution in the class and it's tests. 

The unit test distribution function validation values are from the equivalent functions in R.

### DIFF
--- a/src/Numerics/Distributions/Logistic.cs
+++ b/src/Numerics/Distributions/Logistic.cs
@@ -1,0 +1,515 @@
+﻿// <copyright file="Logistic.cs" company="Math.NET">
+// Math.NET Numerics, part of the Math.NET Project
+// http://numerics.mathdotnet.com
+// http://github.com/mathnet/mathnet-numerics
+//
+// Copyright (c) 2009-2015 Math.NET
+//
+// Permission is hereby granted, free of charge, to any person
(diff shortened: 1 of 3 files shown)


cdrnet (approved):

cdrnet:
Ah, this has been asked for multiple times already - thanks a lot for this contribution!

src/Numerics/Distributions/Logistic.cs
```diff
@@ -0,0 +1,514 @@
...
+
+        /// <summary>
+        /// Gets the standard deviation (σ) of the logistic distribution. Range: σ > 0.
+        /// </summary>
+        public double StdDev => Math.Sqrt(Variance);
+
+        /// <summary>
+        /// Gets the variance of the logistic distribution.
+        /// </summary>
+        public double Variance => (Math.Pow(_scale, 2) * Math.Pow(Math.PI,2))/3;
```
SunnyWar, line 193: I think it's more efficient in this case to use __scale * _scale * Math.PI * Math.PI / 3;
or _scale * _scale * 3.289868133696452872944;

PR #757, opened 2021-03-05
Minor performance

febkor:
Couple of minor changes, mainly for performance, given how (relatively) slow Math.Pow() tends to be.
Might be worth having a seperate optimized class for common power operators like Pow2(), Pow10(), Square(), Cube()


cdrnet (approved):

cdrnet:
Thanks a lot!

> Might be worth having a separate optimized class for common power operators like Pow2(), Pow10(), Square(), Cube()

Yes. Maybe also Pow10 with positive powers.

Issue #672, opened 2019-11-11
NonlinearMinimizerBase and derived classes: static properties overwritten by instances


ibauersachs:
The class `NonlinearMinimizerBase` has various static properties that are initialized/overwritten from instance constructors and methods. This prevents concurrent usage of any of the derived algorithms in parallel and is a _very_ confusing API. E.g. the following does NOT work, the two calls to `FindMinimum` interfere with each other, and `initialMu` is also overwritten.

```
var lm1 = new LevenbergMarquardtMinimizer(initialMu: 0.01);
var lm2 = new LevenbergMarquardtMinimizer(initialMu: 0.02);
var model = ObjectiveFunction.NonlinearModel(..., x, y);

var scales1 = new[] {.1, 1, 10, 10};
var scales2 = new[] {.1, 1, 10, 10, 10, 10};
var start1 = new[] {1.0, 0.0, 0.0, 0.0};
var start2 = new[] {1.0, 0.0, 0.0, 0.0, 0.0, 0.0};

Task.WaitAll(
   Task.Run(() => lm1.FindMinimum(model, start1, null, null, scales1)),
   Task.Run(() => lm2.FindMinimum(model, start2, null, null, scales2))
);
```

Commit 19bd5f576da4c75d0c1a088efc29d9ca9cee4356 introduced these changes and I don't understand why this was designed as it is.

xantiva:
Good point. I started working with Math.Lab and have like to fit e.g. >1.500 curves as parallel as possible. Therefore the library has to be thread safe ... 
But I can find [only questions about thread safety](https://discuss.mathdotnet.com/search?q=thread%20sa%20category%3A5), no answers (only for Fit.Line from 2016)?

keithyipkw:
They are not necessarily about thread safety, e.g., nested optimizations in total least squares regression (of course you can work around this by doing a save and load of the variables). What is the rationale behind using `static` variables?

Commit 19bd5f5 was a refactor. The API has been like this since the beginning in commit aca34e13.

cdrnet:
These properties are no longer static, starting with upcoming v5. Thanks for bringing this up, and thanks @diluculo for the PR!

pzahra:
I ended up making a similar fix on our local copy a few weeks ago, is there an estimation on when v5 is coming?

cdrnet:
No, but there are pre-release packages already available on NuGet.

PR #762, opened 2021-05-02
Remove static properties in LevenbergMarquardtMinimizer

diluculo:
Please review this. Refer to #672.


cdrnet:
Thanks - yes, I think we can and should change this, but this will be a breaking change, so we'll need to align this with a major version increment. Luckily I planned to do that soon anyway for another reason (droping support for some old TFMs).

PR #772, opened 2021-06-30
add JsonInclude for Polynomial and DescriptiveStatistics

### LINKED ISSUE #771: System.Text.Json cannot deserialize Polynomial coefficients correctly
Since the coefficients have a private setter than System.Text.Json cannot correctly deserialize the object.

Either the setter can be made public or add JsonInclude attribute.

Would you be open to include support for this in NET5?

```csharp
#if NET5_0_OR_GREATER
using System.Text.Json.Serialization;
#endif

namespace MathNet.Numerics
{

    public class Polynomial : IFormattable, IEquatable<Polynomial>, ICloneable
    {
        /// <summary>
        /// The coefficients of the polynomial in a
        /// </summary>
        [DataMember(Order = 1)]
#if NET5_0_OR_GREATER
        [JsonInclude]
#endif
        public double[] Coefficients { get; private set; }
```

I can submit a PR if you agree with the solution :)

jetersen:
fixes #771

### DIFF
--- a/src/Numerics.Tests/PolynomialTests.cs
+++ b/src/Numerics.Tests/PolynomialTests.cs
@@ -444,7 +444,7 @@ namespace MathNet.Numerics.UnitTests
             }
 
             // static Complex Evaluate(Complex, double[])
-            {                
+            {
                 var actual = Polynomial.Evaluate(zComplex, c);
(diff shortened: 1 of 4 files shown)


cdrnet:
Perfect, thanks a lot!

Issue #510, opened 2017-08-29
BfgsBMinimizer - ArgumentOutOfRangeException


rostov-da:
Hello. There is a problem with the BfgsBMinimizer  (version 3.20.0.0)
ArgumentOutOfRangeException, 
QuadraticGradientProjectionSearch.cs,  
string number 86.

```
using System;
using MathNet.Numerics.LinearAlgebra;
using MathNet.Numerics.LinearAlgebra.Double;
using MathNet.Numerics.Optimization;
using MathNet.Numerics.Optimization.ObjectiveFunctions;

namespace TestInterface.MathLibTest
{
    public static class MathNetNumericsTest
    {
        public static void Make()
        {
            Func<Vector<double>, double> function = (Vector<double> vectorArg) =>
            {
                double x = vectorArg[0];
                double y = -1.0 * Math.Exp(- x * x);
                return y;
            };
            
            double gradientTolerance = 1e-10;
            double parameterTolerance = 1e-10;
            double functionProgressTolerance = 1e-10;
            int maxIterations = 1000;

            var initialGuess = new DenseVector(new[] { -1.0 });

            // Bad Bound
            var lowerBound = new DenseVector(new[] { -2.0 });
            var upperBound = new DenseVector(new[] { 2.0 });

            // Good Bound
            //var lowerBound = new DenseVector(new[] { -3.0 });
            //var upperBound = new DenseVector(new[] { 3.0 });

            var objective = ObjectiveFunction.Value(function);
            var objectiveWithGradient = new ForwardDifferenceGradientObjectiveFunction(objective, lowerBound, upperBound);
            var algorithm = new BfgsBMinimizer(gradientTolerance, parameterTolerance, functionProgressTolerance, maxIterations);
            var result = algorithm.FindMinimum(objectiveWithGradient, lowerBound, upperBound, initialGuess);

            var resultVector = result.MinimizingPoint;
            double xResult = resultVector[0];
        }
    }
 #}
```

  
#

Shiney:
Looks like the issue is that the loop on line 81 of QuadraticGradientProjectionSearch.cs, will never work if there is only one parameter, this can probably be fixed by making the check  for `jj + 1 >= orderedBreakpoint.Count - 1)` happen at the start of the loop instead

Shiney:
FYI Going to try to create a fix for this

Shiney:
[...]

PR #858, opened 2021-08-20
Fix issue 510

Shiney:
Fixed an index out of bounds issue when calculating BFGS minimizer with one variable.

Moved a check to the start of a loop in QuadraticGradientProjectionSearch so that we exit earlier with one variable, and added a test that would have caught this issue.

Fixes github issue #510


Shiney:
Fixes #510

Shiney:
I don't think I have permissions to link this to the right issue but it is linked to one

cdrnet:
Thanks a lot for investigating this problem! The change looks good to me.

PR #880, opened 2021-11-13
Span<T> & Memory<T> on Fourier

macaba:
An example how an updated API might look (and naive fixing of unit tests). 

Whilst doing this, I saw many opportunities for reducing memory allocations so this would probably be the subject of a 2nd pass through this.


macaba:
Pain points:

The CommonParallel.For and Storage<T> APIs are not Span compatible, hence the fallback to Memory<T> and .ToArray(). Hopefully someone more familiar with Math.Net can weigh in here.

cdrnet:
Thank you!

I really would like to move forward in this direction, and also have some thoughts already. I'll try to get back to you as soon as time permits!

PR #882, opened 2021-11-29
add Marcum-Q function

delicioustuna:
Added the Marcum-Q function to SpecialFunctions and its tests to SpecialFunctionsTests.
The paper and the original implementation code can be found at the following URL
https://dl.acm.org/doi/abs/10.1145/2591004


cdrnet:
This looks excellent, thank you!

Some of the included helper functions may also be interesting on their own. There may also be some overlap with functions we already have, but we'd have to verify equivalence first before trying to bring them together. So I'm fine with including them internally, in isolation, as implemented.

PR #886, opened 2021-12-09
Match the behavior of SystemRandomSource to System.Random

hikarin522:
In `.NET 6`, `new System.Random()` with no arguments is changed to `xoshiro256**`.
https://github.com/dotnet/runtime/pull/47085

### DIFF
--- a/src/Numerics/Random/SystemRandomSource.cs
+++ b/src/Numerics/Random/SystemRandomSource.cs
@@ -49,16 +49,18 @@ namespace MathNet.Numerics.Random
         /// <summary>
         /// Construct a new random number generator with a random seed.
         /// </summary>
-        public SystemRandomSource() : this(RandomSeed.Robust())
+        public SystemRandomSource()
         {
+            _random = new System.Random();
         }
 
         /// <summary>
         /// Construct a new random number generator with random seed.
         /// </summary>
         /// <param name="threadSafe">if set to <c>true</c> , the class is thread safe.</param>
-        public SystemRandomSource(bool threadSafe) : this(RandomSeed.Robust(), threadSafe)
+        public SystemRandomSource(bool threadSafe) : base(threadSafe)
         {
+            _random = new System.Random();
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


cdrnet:
Related, doesn't .NET 6 also introduce a shared Random instance?

cdrnet:
Note, the reason we pass in a seed is to avoid the problem Random used to have when you create multiple instances very quickly they are not guaranteed to be independent (they could end up with the same seed). Is this also resolved then in .NET 6?

hikarin522:
> Related: doesn't .NET 6 also introduce a shared Random instance?

When # 885 is merged, I will issue the following commit pull request.

https://github.com/hikarin522/mathnet-numerics/commit/97094efa85eeefa2403f2fb503bad1905e9c2601

hikarin522:
> Note, the reason we pass in a seed is to avoid the problem Random used to have when you create multiple instances very quickly they are not guaranteed to be independent (they could end up with the same seed). Is this also resolved then in .NET 6?

https://docs.microsoft.com/dotnet/api/system.random.-ctor#System_Random__ctor

> In .NET Framework, the default seed value is derived from the system clock, which has finite resolution. As a result, different Random objects that are created in close succession by a call to the parameterless constructor have identical default seed values and, therefore, produce identical sets of random numbers. You can avoid this problem by using a single Random object to generate all random numbers. You can also work around it by generating your own random seed value and passing it to the Random(Int32) constructor. For more information, see the Random(Int32) constructor.
In .NET Core, the default seed value is produced by the thread-static, pseudo-random number generator, so the previously described limitation does not apply. Different Random objects created in close succession produce different sets of random numbers in .NET Core.

The `.NET Framework` causes problems so I'll fix it.

hikarin522:
Fixed.

cdrnet:
I have not merged yet because I first would like to bring out v5 which mainly changes a few things around native providers and how they are isolated, loaded an packaged (and make them work again for Linux if possible). As soon as this is out, we can finally start to move towards .NET 5 and 6. Sorry for this delay.

PR #887, opened 2021-12-11
add log1p function

delicioustuna:
Added log1p function and its test to SpecialFunctions.
The comparison values used in the test refer to the results of julia.

### DIFF
--- a/src/Numerics.Tests/SpecialFunctionsTests/SpecialFunctionsTests.cs
+++ b/src/Numerics.Tests/SpecialFunctionsTests/SpecialFunctionsTests.cs
@@ -311,5 +311,26 @@ namespace MathNet.Numerics.UnitTests.SpecialFunctionsTests
         {
             AssertHelpers.AlmostEqualRelative(x, SpecialFunctions.Logistic(p), 14);
         }
+
+        /// <summary>
+        /// Log1p function.
+        /// </summary>
+        /// <param name="y">Expected value.</param>
+        /// <param name="x">Input X value.</param>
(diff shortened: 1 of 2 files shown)


cdrnet:
Great, thank you! We can bring this already in for v5.

Btw, do you by chance know whether there is also a more standard short name for that ExponentialMinusOne function we also have (and which is somewhat related to log1p)?

NetPro69:
Hi, no idea. But I didn't do that log1 función. I added the method that
returns the points where the cubic spline has derivative 0 (get horizontal
points) and also the get max point which uses the derivative 0 points.

Kind regards
Edd

delicioustuna:
Thank you for your comments. I am glad.
As far as I know, "Expm1" is used in similar functions in other languages.
For example, Python provides it as "scipy.special.expm1", and c++ provides it as "std::expm1".
[scipy.special.expm1](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.expm1.html)
[std::expm1](https://en.cppreference.com/w/cpp/numeric/math/expm1)

delicioustuna:
> Btw, do you by chance know whether there is also a more standard short name for that ExponentialMinusOne function we also have (and which is somewhat related to log1p)?

Thank you for your comments. I am glad.
As far as I know, "Expm1" is used in similar functions in other languages.
For example, Python provides it as "scipy.special.expm1", and c++ provides it as "std::expm1".
[scipy.special.expm1](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.expm1.html)
[std::expm1](https://en.cppreference.com/w/cpp/numeric/math/expm1)

PR #888, opened 2021-12-15
Added function go get points where cubic spline derivative is 0 and Get Max value for cubic spline in domain

NetPro69:


cdrnet:
That looks very useful, thanks you!

Maybe we could at the same time also evaluate the minimum points (e.g. as value tuple)?

Then, for me points refers to a x-y tuple, not just an x or a y value. I wonder whether it would make sense for both functions to also return the corresponding values? Maybe again as x,y value tuples? Would that make this more useful, or rather more cumbersome to use?

NetPro69:
Yes. I can add that. I will do the commit later today, i am not at my desk

NetPro69:
Hi: I changed the names of the functions so they are more meaningful, people can just interpolate on the x values returned if needed. The GetMinMaxTValue returns a tuple with item1 being the min x or t value and item2 being the max. I also added unit tests

NetPro69:
Is anything missing here to get this feature into the next release?

src/Numerics/Interpolation/CubicSpline.cs
```diff
@@ -623,5 +623,87 @@ int LeftSegmentIndex(double t)
...
+                }
+            }
+            return points.ToArray();
+        }
+
+        /// <summary>
+        /// Returns the t values in the domain of the spline for which it takes the minimum and maximum value.
+        /// </summary>
+        /// <returns>A tuple containing the t value for which the spline is minimum in the first component and maximum in the second component </returns>
+        public Tuple<double, double> GetMinMaxTValues()
```
febkor, line 670: The extrema of a differentiable function on a closed interval will be at its critical points or at the edges of the interval. Not sure the each input (x, y) point has to be tested.

For eg, something like this should be sufficient to get the co-ord of the minimum:
```csharp
var xs = CriticalPoints().Append(_x[0]).Append(_x[^1]);
var ys = xs.Select(x => Interpolate(x));
var minCoord = xs.Zip(ys, (x, y) => (x, y)).MinBy(p => p.y);
```

Given how simple the above code is, maybe this function might not be necessary?
If you do decide to keep it, then `Extrema()` might be a better name.

cdrnet:
Rebased on master to fix the unrelated build issues

cdrnet:
Merged - thank you!

PR #893, opened 2022-01-07
RandomExtensions.NextBigIntegerSequence

Silver-Fang:
I'd like to add support for System.Numerics.BigInteger number type on random generation

### DIFF
--- a/src/Numerics.Tests/Random/RandomExtensionTests.cs
+++ b/src/Numerics.Tests/Random/RandomExtensionTests.cs
@@ -29,6 +29,8 @@
 
 using MathNet.Numerics.Random;
 using NUnit.Framework;
+using System.Numerics;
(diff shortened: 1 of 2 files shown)


Silver-Fang:
I checked the current master code and found the code change wasn't actually applied. What happened? @cdrnet

cdrnet:
Oh, weird - it looks like you reverted it again in #894 which was (accidentally?) opened from master instead of from a feature branch?

cdrnet:
I can revert the revert :)

Silver-Fang:
> I can revert the revert :)

Yes. I accidentally commit the function extension to the bug fix branch so I reverted it. Now this is the side effect. Revert the revert it please. 😀

cdrnet:
Should be back: e7bcab1

PR #895, opened 2022-01-18
Fix for COO sparse matrices not correctly removing duplicates

mohamedmoussa89:
This is my first ever pull request so hopefully I am doing things correctly.

I noticed when writing a solver for beam section property calculations that I was getting incorrect results.  I have tracked this down to incorrect removal of duplicates in `NormalizeDuplicates` for COO matrices.

The fix is simple, see comment on c117e10.  The first commit adds a failing test, which is then fixed so it passes.  

As part of adding the test, I also added two matrix data files (one with and one without duplicates, the end result should be the same).  I wrote my own code to read the COO format, and this lives in the Test project.  I couldn't find any other code in the project to read COO matrices.


cdrnet:
Thanks a lot for the analysis, the good explanation and especially also for providing tests with reference data. The pull request is perfect, thank you!

cdrnet:
Good point regarding reading the COO format. Historically this would have been something for the Data.Text library, although I think it was a mistake to put this into a separate library.

PR #901, opened 2022-02-01
Fix infinite loop in RandomExtensions.NextBigIntegerSequence(0,1)

Silver-Fang:
I found a boundary condition that falls into an infinite loop. Fixed.


Silver-Fang:
@cdrnet AppVeyor has some problem with .Net 6.0?

cdrnet:
Yes, unfortunately the build system is in the middle of a migration, a state which was rather brittle and now apparently broken (although still works on my machine for some reason). I need to finish the migration to fix this properly. At least afterwards it will be technologically simpler, and rely more on the known dotnet cli tools.

cdrnet:
That seems to have worked - master is green again :)

cdrnet (approved):
Thank you!

PR #906, opened 2022-02-14
Combinatorics.GenerateVariation for large N

Silver-Fang:
The space complexity of Fisher-Yates Shuffling is O(n+k). When N is very large, the algorithm will be unexecutable in limited memory, and a more memory-efficient algorithm is needed.


cdrnet:
Very interesting, thank you!

Issue #776, opened 2021-07-10
Inconsistency with PointwiseMultiplication for SparseMatrix


ScottAlexanderCameron:
In-place pointwise multiplication produces different results for `DenseMatrix` than it does for `SparseMatrix`.
example:
```c#
Matrix x = SparseMatrix.OfDiagonalArray(new double[] {1, 2, 3, 2});
Matrix y = DenseMatrix.OfDiagonalArray(new double[] {2, 1, 3, 2});
y.PointwiseMultiply(x, y);
Console.WriteLine(y);
```
produces the following output
```
DenseMatrix 4x4-Double
2  0  0  0
0  2  0  0
0  0  9  0
0  0  0  4
```

while for `SparseMatrix`, the output of the following code
```c#
Matrix x = SparseMatrix.OfDiagonalArray(new double[] {1, 2, 3, 2});
Matrix y = DenseMatrix.OfDiagonalArray(new double[] {2, 1, 3, 2});
x.PointwiseMultiply(y, x);
Console.WriteLine(x);
```
is
```
SparseMatrix 4x4-Double 0.00 % Filled
0  0  0  0
0  0  0  0
0  0  0  0
0  0  0  0
```

I have only tested this with matrices of doubles.
The root of this seems to be in the implementation of `DoPointwiseMultiply()` which calls `Clear()` on the result without checking whether or not the result matrix is one of the operands.

One would expect that the results are the same for `SparseMatrix` and `DenseMatrix` or at least an exception be thrown if this behaviour is not supported.

jkalias:
Should be fixed in PR #912

cdrnet:
Released in 5.0.0-beta2 - thank you both for reporting and for the patch!

PR #912, opened 2022-03-25
Pointwise multiplication and division on sparse matrices

jkalias:
The result of pointwise multiplication and division on sparse matrices is passed as an argument. The result is cleared without checking if it is equal to the current instance, thus causing a full deletion of the sparse matrix contents.
 (Fixes bug "Inconsistency with PointwiseMultiplication for SparseMatrix #776")


cdrnet:
Great, thank you!

jkalias:
Can you please update the contributors list? 🤓

cdrnet:
Yes, of course - just updated. I should automate this sometime.

jkalias:
Great, thank you!

Issue #926, opened 2022-05-24
NonlinearObjectiveFunction.SetObserved method throws overflow exception for large weighted vectors


laidBackBird:
I'm using the `NonlinearObjectiveModel` for nonlinear least squares fitting of weighted data. However, the fitting fails due to an overflow exception when the number of data points becomes too large (let's say, 10000+). 

The origin of this exception is the `SetObserved` method in the [`NonlinearObjectiveFunction`](src/Numerics/Optimization/ObjectiveFunctions/NonlinearObjectiveFunction.cs). At the end of this method, a `DenseMatrix` is constructed for the data point weights, although this matrix contains diagonal elements only. The dense matrix stores all (10000+)^2 (mostly zero) entries, which exceeds my memory. To my mind, a `DiagonalMatrix` would be the better choice here.

I've tested replacing line 217-219 with
```
Weights = (weights == null)
    ? null
    : Matrix<double>.Build.DiagonalOfDiagonalVector(weights);
```
... and it solves my issue. I also tested this for small data (10 points) and didn't notice any performance declines there either. 

Is there any reason for hanging on to the `DenseMatrix`? If not, I would suggest modifying the mentioned lines as described. I'm happy to great a pull request for this.

diluculo:
@laidBackBird, you are right.

Actually, the `Weightings` matrix is only used to decompose into `W = LL'`, i.e. `L = Weights.Diagonal().PointwiseSqrt()`

laidBackBird:
@diluculo That's true, I didn't follow it further. Storing `L` should be sufficient, since the `Weights` apparently aren't used otherwise. I could imagine that keeping them was a "consistency choice"?!

How should we proceed? Should I implement the (minor) changes and create a pull request? If so, would you suggest removing the `Weights` property from the class?

diluculo:
@laidBackBird, the 'Weights' is not only set initially but can be changed at each iteration step in some cases. That is, it needs to be verified later. It is why I made 'Weights' public property.

PR #927, opened 2022-05-30
Feature/fix overflow exception when fitting large data

laidBackBird:
Fixes mathnet/mathnet-numerics#926


Crown0815:
@cdrnet This fix resolves a major issue with fitting larger datasets and seems to be minimally invasive. Is there any reason this has not been merged yet?

PR #932, opened 2022-07-04
Feature/allow mixture of bounded and unbounded fit parameters

### LINKED ISSUE #931: Nonlinear minimizers don't allow mixing of bounded and unbounded parameters
Fit model parameters may optionally be box constrained for nonlinear minimization (see e.g. [NonlinearMinimizerBase.cs](https://github.com/mathnet/mathnet-numerics/blob/834a2d6425b66042d7528154e2e2532a75540b04/src/Numerics/Optimization/NonlinearMinimizerBase.cs)). However, a current limitation is that one has to either provide a full set of lower and/or upper bounds, or leave all parameters unbounded (in any direction). 

Sometimes it makes sense for some parameters to be bounded, e.g. due to physical constraints, while it doesn't make sense for others. It would be great if it was possible to mix bounded and unbounded parameters. This could, for instance, be implemented by allowing individual bounds to be nullable, and then checking/catching the individual "null-bounds" in the `ProjectToInternalParameters` method.

AutWint:
No tests added (so far).

Fixes #931

### DIFF
--- a/src/Numerics/Optimization/NonlinearMinimizerBase.cs
+++ b/src/Numerics/Optimization/NonlinearMinimizerBase.cs
@@ -1,5 +1,6 @@
 ﻿using MathNet.Numerics.LinearAlgebra;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 
 namespace MathNet.Numerics.Optimization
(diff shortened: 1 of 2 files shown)


Crown0815:
We are currently using a workaround to get this feature to work. @cdrnet Any chance of getting this integrated the near future?

PR #938, opened 2022-08-15
Extended matlab parser with structures, cell matrices and character matrices

jan-van-dieren:
Since https://discuss.mathdotnet.com/c/numerics is giving me a timeout I would like to use this PR to ask the Mathnet maintainers if there is a chance for this parsing code to be included.

What can it do?

- Parse cell matrices both at the file root and when nested
- Parse structures both at the file root and when nested
- Parse character matrices (Only UTF8 tested right now)
- Parsing regular numeric matrices is unchanged, also when they are inside a cell or structure

What still needs to be done for PR readiness (as far as im aware)

- More commenting
- Tests for different char array encodings
- Fixing some code duplication between cell matrix and structure parsing


jan-van-dieren:
This is all based on the V5 spec at https://www.mathworks.com/help/pdf_doc/matlab/matfile_format.pdf

cdrnet:
Thank you very much!

PR #956, opened 2022-09-21
Fix initialisation bug in sparse matrix DoTransposeThisAndMultiply operation

rca22:
The result vector needs to be zeroed before contributions from the non-zero elements of the sparse matrix are added to it.


rca22:
Fixes #957

cdrnet:
Thank you!

PR #959, opened 2022-09-28
Robust Newton Raphson root on the bound

jkalias:
In the degenerate case for which the function was zero everywhere, `RobustNewtonRaphson` returned the lower bound as a valid solution.

Though this is mathematically correct and all roots are valid, a more practical root corresponds to the mean value of the search range.

In addition, this PR addresses a small inconsistency, for which `fx` was checked directly against 0.0, thus ignoring the given accuracy.


jkalias:
Hi @cdrnet. What do you think of this PR?

Arlofin:
Some technical comments:

1. The proposal produces an overflow for `lowerBound = Double.MaxValue - 2` and `upperBound = Double.MaxValue - 1`. The technically correct way to compute the mid-point between finite numbers a and b is a + 0.5 * (b - a). The issue was there before, but it would be an opportunity to improve.
2. If lower bound is -∞ or upper bound is +∞, the proposal does not produce useful results.
3. What should be the behaviour if lower bound is -∞ and upper bound is +∞? Mathematically, this is a useful specification, but the implementation is not able to cope with it. The issue was there before, but, again, it would be an opportunity to improve. I can imagine two reasonable ways to handle it: (a) Choose an arbitrary finite start value, e.g. 0 (b) Throw an `ArgumentOutOfRangeException` (which should also be done if any NaN is input).
4. In the proposal, `fmin` and `fmax` are not used in the first case (`Math.Abs(fx) < accuracy`), so the computation should be done afterwards only (the compiler might optimize this anyway, but who knows).

jkalias:
1. Agree, will fix.
2. If the boundaries are +/- inf, essentially this is unconstrained Newton search. In such a case, I can't really tell what a sensible starting point would be, though 0.0 seems a reasonable start.
4. I don't really understand what you mean here. Could you please explain in more detail?

Arlofin:
The difference between 2 and 3 (sorry, if this was not clear) is that 2 refers to one-sided constraint, like finding a root in [0, ∞[, and 3 refers to unconstrained root finding. It is mostly a question whether there should be support for these cases or not. Clearly, 3 is more difficult due to the lack of natural initial value, therefore I separated the two questions. Furthermore I believe, admittedly without having tested it, that one-sided constraints with your proposal never work while previously there were some chances for convergence, so I wanted to point out the difference in behaviour.
After some further thought, I think it would be in the spirit of the method to not support any infinite constraint, although the problem is mathematically well-specified and the Newton method does not need constraints (although they are helpful to check out of bounds) and hence convergence might occur. However, the 'robust' part of the method relies on bisection, which needs finite constraints. If we can agree on this, my wish would be that the deliberate lack of support be made transparent in the comment, and by throwing an `ArgumentOutOfRangeException`.
Note again, that this has not much to do with your actual proposal, but it fits well to the title of your PR, hence I saw some opportunities.

jkalias:
The suggestions are implemented, thanks for pointing these out.Any other problems that someone else sees which slipped my attention?

jkalias:
Any news regarding this PR?

jkalias:
Hi @cdrnet, I understand that this is a side project for you, and that you have no obligation to reply to any of my PR messages. I'd really appreciate it if you could have a look at this.

cdrnet:
@jkalias thank you - yes, I'll have a look

cdrnet:
Thank you - yes, I'm fine with the change to first test the mean, and agree that explicitly rejecting infinity is ok for the "robust" variant. Thank you also for adding tests!

Just a small thing before I can merge: it looks like the PR also includes some changes in the project files which shouldn't be there (and which now also conflict). This could be resolved e.g. by rebasing your branch on current master and then skipping the project files.

cdrnet (changes requested):
Once the project file changes are in I can merge - thanks again!

src/FSharp.Tests/FSharp.Tests.fsproj
```diff
@@ -34,6 +34,8 @@
         <Compile Include="Main.fs" />
     </ItemGroup>
     <ItemGroup>
+        <PackageReference Include="FSharp.Core" Version="6.0.6" />
+        <PackageReference Include="System.ValueTuple" Version="4.5.0" Condition="'$(TargetFramework)' == 'net461'" />
```
cdrnet, line 38: please drop these again (they likely got in through some automatic merging)
```suggestion
```

src/FSharp/FSharp.fsproj
```diff
@@ -68,6 +68,8 @@ Control.Describe now includes CPU architecture and family identifier if know</Pa
         <Compile Include="Quaternion.fs" />
     </ItemGroup>
     <ItemGroup>
+        <PackageReference Include="FSharp.Core" Version="6.0.6" />
+        <PackageReference Include="System.ValueTuple" Version="4.5.0" Condition="'$(TargetFramework)' == 'net461'" />
```
cdrnet, line 72: please drop these again
```suggestion
```

src/Numerics.Tests/Numerics.Tests.CUDA.csproj
```diff
@@ -27,6 +27,8 @@
     <PackageReference Include="NUnit" Version="3.13.3" />
     <PackageReference Include="NUnit3TestAdapter" Version="4.3.0" />
     <PackageReference Include="NUnitLite" Version="3.13.3" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3">
```
cdrnet, line 31: please drop these again
```suggestion
```

src/Numerics.Tests/Numerics.Tests.OpenBLAS.csproj
```diff
@@ -27,6 +27,8 @@
         <PackageReference Include="NUnit" Version="3.13.3" />
         <PackageReference Include="NUnit3TestAdapter" Version="4.3.0" />
         <PackageReference Include="NUnitLite" Version="3.13.3" />
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
+        <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3">
```
cdrnet, line 31: please drop these again
```suggestion
```

jkalias:
Hi @cdrnet , thanks for taking the time. I can't figure out what happens with these automatic bump merges. I think now the PR should be ok, can you please have another look?

src/Numerics/Numerics.csproj
```diff
@@ -52,7 +52,7 @@ Control.Describe now includes CPU architecture and family identifier if know</Pa
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
         </PackageReference>
-        <PackageReference Include="System.ValueTuple" Version="4.4.0" Condition="'$(TargetFramework)' == 'net461'" />
+        <PackageReference Include="System.ValueTuple" Version="4.5.0" Condition="'$(TargetFramework)' == 'net461'" />
```
cdrnet, line 55:
```suggestion
```

ah, missed one

cdrnet, line 55:
```suggestion
        <PackageReference Include="System.ValueTuple" Version="4.4.0" Condition="'$(TargetFramework)' == 'net461'" />
```

ah, missed one

cdrnet (approved):
Looks good now - I'll merge once the build is green. Thanks again!

jkalias:
@cdrnet another question, perhaps off-topic here: why don't we use `var` for all variable declarations in the project, but instead we explicitly write down the type, eg. `double`?

cdrnet:
I prefer to be explicit in that regard here. We don't only offer many algorithms for both single and double precision, but often also for both real and complex numbers. Writing the type explicitly helps avoiding mistakes like accidentally working with the wrong type within some block, and to be more aware when looking at the code whether we may need to treat e.g. real vs complex numbers differently.

cdrnet:
(although it's not a strong concern - I'd also accept code using var, and I think we do use it in a few places)

jkalias:
Ok, thanks for clarifying, and taking the time to merge this PR!

jkalias:
Hi @cdrnet ,

May I suggest that the Release Notes should be updated for this issue? (https://numerics.mathdotnet.com/ReleaseNotes.html)
It would be nice to have the credit mentioned

"Root Finding: Newton-Raphson better handling of zero-evaluations  **~jkalias**"

PR #963, opened 2022-10-03
Explicitly specify invariant culture in some unit tests

Arlofin:
The changes ensure that the unit tests pass on locales that do not use the period as decimal separator


cdrnet:
Thank you!

PR #995, opened 2023-03-13
Add method with option to not allow extrapolation.

Hylke-Atmos:
Hi @cdrnet 

You library is great, however by default the interpolation functions also extrapolate, which is not always desired. Although I know a user is (in most cases) likely able to test possible extrapolation from outside the library, I thought it made a nice addition to the library.

I added a function (to LinearSpline) to which you can provide a flag to indicate whether extrapolation is allowed.

Please let me know if you'd like this addition for the library. It is currently implemented directly in LinearSpline, but I can imagine it might be worth for all interpolation functions. So if you like the idea I can extend it towards the IInterpolation Interface and implement it into all implementations.

Kind Regards,
Hylke

### DIFF
--- a/src/Numerics/Interpolation/LinearSpline.cs
+++ b/src/Numerics/Interpolation/LinearSpline.cs
@@ -130,6 +130,24 @@ namespace MathNet.Numerics.Interpolation
         /// <returns>Interpolated value x(t).</returns>
         public double Interpolate(double t)
         {
+            return Interpolate(t, true);
+        }
+
+        /// <summary>
+        /// Interpolate at point t.
+        /// </summary>
+        /// <param name="t">Point t to interpolate at.</param>
+        /// <param name="allowExtrapolate">set to fals if extrapolate should not be allowed</param>
+        /// <returns>Interpolated value x(t).</returns>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when t is outside range and would result in Extrapolation!</exception>
+        public double Interpolate(double t, bool allowExtrapolate)
+        {
+            if (!allowExtrapolate &&
+                (t < _x[0] || t > _x[_x.Length - 1]))
+            {
+                throw new ArgumentOutOfRangeException("'t' is outside of the constructed array. This would result in Extrapolation!");
+            }
+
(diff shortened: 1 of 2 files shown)


src/Numerics/Interpolation/LinearSpline.cs
```diff
@@ -130,6 +130,24 @@ public static LinearSpline Interpolate(IEnumerable<double> x, IEnumerable<double
...
+        }
+
+        /// <summary>
+        /// Interpolate at point t.
+        /// </summary>
+        /// <param name="t">Point t to interpolate at.</param>
+        /// <param name="allowExtrapolate">set to fals if extrapolate should not be allowed</param>
+        /// <returns>Interpolated value x(t).</returns>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when t is outside range and would result in Extrapolation!</exception>
+        public double Interpolate(double t, bool allowExtrapolate)
```
jkalias, line 143: I am not a big fan of ```bool``` variables in public APIs which act like toggle switches. I would rather create another public function, something like ```InterpolateWithPossibleExtrapolation```, which probably needs to defined on the interface, so every instance can implement it.

jkalias: Hi again. I am in favor of having a clean API, and I feel that your proposal is in the right direction.

You are right, the name I suggested should have been ``` InterpolateWithoutExtrapolation ```

febkor, line 143: That all the interpolators extrapolate by default is possibly surprising behaviour, especially given that is not documented (as far as I can tell).
But I wonder if `InterpolateWithoutExtrapolation` or the proposed implementation (with `allowExtrapolate`) will resolve this in general.
Presumably users of the other splines would also want this method which would mean adding this method to all classes and to the interface.

Perhaps a cleaner implementation would be to create a decorator class called `NonExtrapolatingSpline` (i.e., implements `IInterpolation` and wraps `IInterpolation`), that handles this in general for all the interpolators.
Also, maybe doc comments could be added to each of the interpolators indicating that they extrapolate by default, but that this new class exists if anyone wants to opt out.

PR #999, opened 2023-03-24
Bug in string description for mx0 and 0xn matrices

### LINKED ISSUE #775: Bugs with 1x0 and 0x1 matrices
I find that there are several bugs with 1x0 and 0x1 matrices (Matrix<double>.Dense).

- The method toString() cannot work normally with 1x0 and 0x1 matrices (althogh it does work with 0x0 matrices)
- When computing AxB (A is 1x0 matrix and B is 0x1 matrix) with MKL library, there will be a ERROR: "Intel MKL ERROR: Parameter 9 was incorrect on entry to cblas_dgemm", which severely affects the computation speed.

BTW, such matrices are of practical application. For example, in state space model, we need 0x0, 0x1 and 1x0 matrices to describe a gain-only process.

jkalias:
Fixes crashes for matrices whose dimension can be zero (fixes #775 )


cdrnet:
Thank you!

PR #1000, opened 2023-03-24
Bug fix: no crash for zero vector in ResidualStopCriterion.

### LINKED ISSUE #720: [BUG] ResidualStopCriterion fails if iterative solver is solved for zero vector.
https://github.com/mathnet/mathnet-numerics/blob/d96a65fb21da0777888ed5f2824237399801cb49/src/Numerics/LinearAlgebra/Solvers/ResidualStopCriterion.cs#L198

The comparision should actually be "less or equal". The following code will throw a `NumericalBreakdownException`:
```c#
var A = new DenseMatrix(2, 2, new double[] { 1, 2, 3, 4 });
var b = new DenseVector(2); // zero vector!
var x = new DenseVector(2);

var iterator = new Iterator<double>(new ResidualStopCriterion<double>(1e-5));
var solver = new BiCgStab();

solver.Solve(A, b, x, iterator, null);
```

jkalias:
Fixes #720


cdrnet:
Thank you!

PR #1013, opened 2023-06-01
Integration: Add triple integral

diluculo:
I've recently encountered the need to perform triple integration over tetrahedral regions. To address this, I've added a very simple yet convenient 3D integral method to the `GaussLegendreRule` class.

Regarding the method name, since the 2D integration method is called `OnRectangle`, I've named the 3D integration method `OnCuboid`. However, I'm uncertain if this name is appropriate.


cdrnet:
Thank you!

PR #1017, opened 2023-06-21
Introduce Deming/Orthogonal regression

JesperGr:
Deming/Orthogonal regression, least-Squares fitting the points in the 2D dataset (x,y) to a line
`
   a*x + b*y + c = 0
`
For delta equal 1 (the default value), this is performing orthogonal regression, minimizing the sum of squared perpendicular distances from the data points to the regression line.

Discussed/suggested here:
https://github.com/mathnet/mathnet-numerics/discussions/786

Questions: 
* Is `Fit.Line2D` a good name?
* Where to put `StandardLineToYxLine`?

### DIFF
--- a/src/Numerics/LinearRegression/DemingRegression.cs
+++ b/src/Numerics/LinearRegression/DemingRegression.cs
@@ -1,0 +1,203 @@
+﻿// <copyright file="DemingRegression.cs" company="Math.NET">
+// Math.NET Numerics, part of the Math.NET Project
+// http://numerics.mathdotnet.com
+// http://github.com/mathnet/mathnet-numerics
+//
+// Copyright (c) 2009-2023 Math.NET
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+// OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+// HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+// WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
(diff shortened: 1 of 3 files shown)


jkalias:
> * Is `Fit.Line2D` a good name?
I think that `Fit.Deming` would be more appropriate. `Line2D` seems like it performs the "usual" least squares fitting

> * Where to put `StandardLineToYxLine`?
This is used purely for testing purposes, I think it is more appropriate in the testing class `FitTests.StandardLineToYxLine `

src/Numerics/LinearRegression/DemingRegression.cs
```diff
@@ -0,0 +1,203 @@
...
+        public static (double A, double B, double C) Fit(double[] x, double[] y, double delta = 1.0)
+        {
+            if (x.Length != y.Length)
+            {
+                throw new ArgumentException($"All sample vectors must have the same length. However, vectors with disagreeing length {x.Length} and {y.Length} have been provided. A sample with index i is given by the value at index i of each provided vector.");
+            }
+
+            if (x.Length <= 1)
+            {
+                throw new ArgumentException($"A regression of the requested order requires at least {2} samples. Only {x.Length} samples have been provided.");
```
jkalias, line 76: 2 doesn't need to be in braces {}

src/Numerics/LinearRegression/DemingRegression.cs
```diff
@@ -0,0 +1,203 @@
...
+        /// The parameters (a,b,c) are scaled such that a and b
+        /// in absolute values are always less than one.
+        /// </para>
+        /// </summary>
+        /// <param name="x">X data</param>
+        /// <param name="y">Y data</param>
+        /// <param name="delta">Ratio of variances of x and y data, var(y)/var(x). Default value is 1</param>
+        /// <returns> returning its best fitting parameters as (a, b, c) tuple.</returns>
+        /// <remarks>
+        /// Solution algorithm as described from:
```
jkalias, line 61: I couldn't find the reference, only a wikipedia [linke](https://en.wikipedia.org/wiki/Deming_regression). Could you provide a pdf link or something similar?

src/Numerics/LinearRegression/DemingRegression.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                // When sxx has a value, it is a vertical line
+                if (delta * sxx > syy && sxx > maxax * 1e-10)
+                    return (0, 1, -my);
+                // when syy has a value, it is a horizontal line
+                else if (syy > maxay * 1e-10)
+                    return (1, 0, -mx);
+                // No line could be calculated
+                else
+                    return (0, 0, 0);
```
jkalias, line 136: I would expect an algorithm which fails to return `double.NaN`s instead of zeros

src/Numerics/LinearRegression/DemingRegression.cs
JesperGr: Fixed also in SimpleRegression.cs

src/Numerics/LinearRegression/DemingRegression.cs
JesperGr: That was the best link I could find.

JesperGr:
I like the Line in the name, because it is very similar to the `SimpleRegression`. But I am fine with `Deming` as well, so I have updated the `Fit.Line2D` to `Fit.Deming`. And moved the `StandardLineToYxLine` to the `FitTests` - I will then just keep my own line tools alive.

PR #1018, opened 2023-06-27
MKL linux build

gabfeldman:
I'm using MathNet with MKL on Linux. I had to make these changes to build with the latest MKL.
The current MKL for Linux library in the nugget package is too old and MathNet doesn't accept it.


cdrnet:
Ah great, thank you very much!

PR #1037, opened 2023-10-12
Impl API for Broyden based on Vector<double> instead of double[]

axtimhaus:
I wondered why the root finding classes do not use vectors for external API, but only arrays. It would help for my use cases to use Vectors, because I can derive from `DenseVector` to make mapping to my actual data structures there.

I just doubled the original methods and made the respective changes for vectors.

Tests will follow. If this is accepted I will do the same for the other algorithms, too.


axtimhaus:
Copied original tests and modified them for vector API.

Issue #1061, opened 2024-02-04
Incorrect condition for stopping the LevenbergMarquardtMinimizer's Minimum() by parameter tolerance.


anotherche:
https://github.com/mathnet/mathnet-numerics/blob/f19641843048df073b80f6ecfcbb229d3258049b/src/Numerics/Optimization/LevenbergMarquardtMinimizer.cs#L161

Why is the vector length (`Pstep.L2Norm()`) compared to the square of the vector length (`P.DotProduct(P)`)? Looks like a bug. Of course, if the parameters were scaled correctly (`|Pi|~1`) this does not cause a problem. But in general, this is not correct.  Wouldn't it be more correct to compare `Pstep.L2Norm()` to `P.L2Norm()`, or their squares?

diluculo:
You are correct,  It's a bug. 
LINE 160: // if ||ΔP|| <= xTol * (||P|| + xTol), found and stop

cdrnet:
Finally merged - thank you!

PR #1066, opened 2024-02-29
Fix LevenbergMarquardtMinimizer stopping condition

diluculo:
addressing issue #1061

### DIFF
--- a/src/Numerics/Optimization/LevenbergMarquardtMinimizer.cs
+++ b/src/Numerics/Optimization/LevenbergMarquardtMinimizer.cs
@@ -158,7 +158,7 @@ namespace MathNet.Numerics.Optimization
                     Pstep = Hessian.Solve(-Gradient);
 
                     // if ||ΔP|| <= xTol * (||P|| + xTol), found and stop
-                    if (Pstep.L2Norm() <= stepTolerance * (stepTolerance + P.DotProduct(P)))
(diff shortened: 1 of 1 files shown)


jkalias (approved):
I would suggest to write a regression test for this bug

diluculo:
@jkalias,  thank you for your prompt response.

While I may not have a concrete test case to demonstrate the issue, it's essential to highlight that the standard Levenberg-Marquardt (LEVM) stopping criterion is `||ΔP|| <= xTol * (||P|| + xTol)`. Please refer to the stopping criteria as defined on page 26 of [Madsen 2004](http://www2.imm.dtu.dk/pubdb/edoc/imm3215.pdf.) 

image

The original buggy code, using `P.DotProduct(P)`, unintentionally produces an early stopping effect. For instance, let's dig into the details of the `BoxBod_LM_Der()` unit test, specifically examining the section where the condition `parameters < upper, scales` is applied.

| Result    | P.DotProduct(P)   | P.L2Norm()        | Expected      |
|-----------|-------------------|-------------------|---------------|
| Iteration | 16                | 20                |               |  
| P[0]      | 213.80940888      | 213.80940889      | 213.80940889  |  
| P[1]      | 0.54723748553    | 0.54723748542     | 0.54723748542 |

cdrnet:
Thank you!

Issue #1067, opened 2024-03-01
Modify NewtonMinimizer so that it forces Hessian to be positive definite


strMikhailPotapenko:
The current Newton's method implementation has no option to force Hessian to be positive definite, which is a common technique to improve convergence. 

In the current implementation the algorithm will not converge to the global minimum if the following conditions are met:

- Hessian is indefinite
- Directional derivative is negative
- Initial guess is close to a local maximum and the desired global minimum (too vaguely described to be very useful condition but there are cases where the above two are true and the algorithm still converges so I thought I would mention this anyway)

This may not be a complete description of cases when forcing positive definiteness is helpful but I ran into such a case and can illustrate with the [Six Camel Hump Function](http://www.sfu.ca/~ssurjano/camel6.html)

Below is a contour plot of this function with areas that satisfy the first two requirements above highlighted in dark blue. I have picked some points out of those areas and confirmed that the NewtonMinimizer implementation does not converge when using those points as initial guesses.

![image](https://github.com/mathnet/mathnet-numerics/assets/149536475/a24eaf1e-5f3c-4aab-8a43-51fcd900c447)

It would be interesting to write a test that iterates through a grid in the domain above to see if those two conditions are necessary/sufficient for non-convergence but I don't currently have the time and it is outside the scope of this issue.

To restate, an option should be added to NewtonMinimzer that forces the Hessian to be Positive Definite.

I will shortly create a pull request with my proposed solution to this issue.

Regards,
Mikhail

strMikhailPotapenko:
Created and linked a pull request to address this issue. The code changes include a unit test which initializes the NewtonMinimizer at point {1, -0.6} (in one of the blue areas above). If no modification of the Hessian is performed then the algorithm converges to the local maximum of {1.109205317132443, -0.76826824818394157}.

PR #1068, opened 2024-03-01
Modified newtons method

strMikhailPotapenko:
My proposed solution to #1067 (which I also submitted).

I added an option to Newton Minimizer which forces the Hessian to be positive definite by reversing negative eigenvalues. This method is described in 

Philip E. Gill, Walter Murray, and Margaret H. Wright, Practical Optimization, 1981, 107–8.

I also added a unit test, FindMinimum_SixHumpCamel_IndefiniteHessian(), to demonstrate the benefit of this modification. If the optional argument is changed to HessianModifiers.None then the test fails.

Please see the issue for further illustration of the use case.

I made the optional argument an enum because this method of forcing positive definiteness can be improved upon in terms of efficiency and this gives some future developer the hooks to be able to do so.

Regards,
Mikhail

### DIFF
--- a/src/Numerics/Optimization/HessianModifiers.cs
+++ b/src/Numerics/Optimization/HessianModifiers.cs
@@ -1,0 +1,43 @@
+﻿// <copyright file="HessianModifiers.cs" company="Math.NET">
+// Math.NET Numerics, part of the Math.NET Project
+// https://numerics.mathdotnet.com
+// https://github.com/mathnet/mathnet-numerics
+//
+// Copyright (c) 2009-$CURRENT_YEAR$ Math.NET
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+// OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+// HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+// WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+// OTHER DEALINGS IN THE SOFTWARE.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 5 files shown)


strMikhailPotapenko:
@cdrnet, Would it be possible to get this pull request reviewed?

strMikhailPotapenko:
@jvangael, @jkalias, @JohanLarsson, I think you are also listed as member of the Math.Net team. Would any of you be willing to consider this pull request?

jkalias:
@strMikhailPotapenko I would be more than happy to. The problem is that I can only merge PRs for the MathNet.Spatial project, for the Numerics only @cdrnet has the relevant rights to.

strMikhailPotapenko:
@jkalias Ah I didn't know that, sorry about that and thank you for the response.

Just looking at open issues and pull requests it does seem like there isn't a lot of activity with this repo, which is too bad because I have been using this library quite often lately and have been pretty satisfied with it. Do you happen to have any information on the plans of @cdrnet as far as actively maintaining this repo or any way to request that information from him?

jkalias:
@strMikhailPotapenko nothing to be sorry about :)

I have no information on whether/when @cdrnet will continue to actively maintain this repo. On the other hand, it's an open source project, so there is no obligation to do so, and rightfully so.

strMikhailPotapenko:
@jkalias Very true!

src/Numerics/Optimization/HessianModifiers.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MathNet.Numerics.Optimization
+{
+    public enum HessianModifiers
+    {
+        None,
+        ReverseNegativeEigenValues
```
jkalias, line 41: I would spell this as `ReverseNegativeEigenvalues`, with a small v

src/Numerics/Optimization/NewtonMinimizer.cs
```diff
@@ -125,6 +128,46 @@ public static MinimizationResult Minimum(IObjectiveFunction objective, Vector<do
...
+            return objective.Hessian.LU().Solve(-objective.Gradient);
+        }
+
+        /// <summary>
+        /// Force the Hessian to be positive definite by reversing the negative eigenvalues. Use the EVD decomposition to then solve for the search direction.
+        /// Implementation based on Philip E. Gill, Walter Murray, and Margaret H. Wright, Practical Optimization, 1981, 107–8.
+        /// </summary>
+        /// <param name="objective"></param>
+        /// <returns></returns>
+        static Vector<double> ReverseNegativeEigenValuesAndSolve(IObjectiveFunction objective)
```
jkalias, line 158: Here also with small v, `ReverseNegativeEigenvaluesAndSolve`

src/Numerics/Optimization/NewtonMinimizer.cs
```diff
@@ -125,6 +128,46 @@ public static MinimizationResult Minimum(IObjectiveFunction objective, Vector<do
...
+        {
+            Evd<double> evd = objective.Hessian.Evd(Symmetricity.Symmetric);
+            for (int i = 0; i < evd.EigenValues.Count; i++)
+            {
+                if (evd.EigenValues[i].Real < double.Epsilon)
+                {
+                    evd.EigenValues[i] = Math.Max(-evd.EigenValues[i].Real, double.Epsilon);
```
jkalias, line 165: You can skip the `if` condition altogether, if you write this as `evd.EigenValues[i] = Math.Max(Math.Abs(evd.EigenValues[i].Real), double.Epsilon);`

strMikhailPotapenko: @jkalias I am sorry I didn't see your very helpful review comments until much later. I accepted your proposal in this comment and all others in the latest commit to this PR. Thank you for taking the time to review!

jkalias (approved):

PR #1093, opened 2024-09-10
Added Differentiate3 to Interpolation

osswaldo:


jkalias:
I would like to see some unit tests for this.

jkalias (approved):

tefracky:
I tool a closer look into the tests and there aren't any tests for the Differentiate methods at all.

osswaldo:
I tool a closer look into the tests and there aren't any tests for the Differentiate methods at all.

jkalias:
Maybe we can start now having tests for it? ☺️

osswaldo:
Ok, done ;)

jkalias (approved):

jkalias:
This is awesome, thank you

src/Numerics/Interpolation/IInterpolation.cs
```diff
@@ -65,6 +65,13 @@ public interface IInterpolation
         /// <returns>Interpolated second derivative at point t.</returns>
         double Differentiate2(double t);
 
+        /// <summary>
+        /// Differentiate three times at point t.
+        /// </summary>
+        /// <param name="t">Point t to interpolate at.</param>
+        /// <returns>Interpolated third derivative at point t.</returns>
+        double Differentiate3(double t);
```
febkor, line 73: Adding a method to IInterpolation is a breaking change since IInterpolation is a public interface that consumers of the library can and do implement.
Might be better to leave it out of the interface but keep the method for the classes  for which the method is supported.

jkalias:
It’s been more than 2 years that the latest 5.0 release of the library has been published. I’m not the official maintainer but I wouldn’t have something against releasing a 5.1 version with this change among others (it’s adding a feature, not removing existing ones or changing functionality)

osswaldo:
It isn't a realy "breaking" change if one extend an interface. You can simply implement the method as not implemented. However, it could be make sence to introduce an IDifferentiableInterpolation in order to remove the SupportsDifferentiation property.

osswaldo:
I am wondering what is the current state of this PR? Will it be merged?

PR #1106, opened 2025-01-03
Change documentation of Svd.Determinant properties

### LINKED ISSUE #1105: Svd.Determinant documentation error
Hello,

This is kind of a trivial thing, but it tripped me up.  The Intellisense for Svd.Determinant says "Gets the determinant of the square matrix for which the SVD was computed."  But it actually returns the absolute value of the determinant.  There are other "Determinant" properties, like Evd.Determinant and  QR.Determinant, which also actually return the absolute value of the determinant, but the Intellisense for them makes that clear.

If you agree that the Intellisense for Svd.Determinant should say "Gets the absolute value of the determinant of the square matrix for which the SVD was computed.", I am happy to submit a PR making that change.

Thanks!

Asa

apacker1:
Fixes #1105

### DIFF
--- a/src/Numerics/LinearAlgebra/Complex/Factorization/Svd.cs
+++ b/src/Numerics/LinearAlgebra/Complex/Factorization/Svd.cs
@@ -89,7 +89,7 @@ namespace MathNet.Numerics.LinearAlgebra.Complex.Factorization
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


jkalias (approved):

cdrnet:
Thank you!

PR #1116, opened 2025-03-21
Enhanced Levenberg-Marquardt Algorithm with Residual Function Support

### LINKED ISSUE #1114: Extend Levenberg-Marquardt Algorithm to Support Direct Function Minimization
I'd like to extend the Levenberg-Marquardt algorithm to support direct function minimization in addition to the existing model fitting approach.

#### Current Model-Based Approach

Currently, the LevenbergMarquardtMinimizer only works with `IObjectiveModel` which represents a curve fitting problem:

- It minimizes the sum of squared residuals: `RSS = Σ w_i * (y_i - f(x_i; p))²`
- Where `f(x; p)` is the model function that predicts `y` given `x` and parameters `p`
- This approach is used when we have observed data points `(x_i, y_i)` and want to find the best model parameters `p`

#### Proposed Function-Based Approach

I'd like to add support for direct objective function minimization:

- Directly minimizes an arbitrary objective function `F(p)` (which can include RSS as a special case)
- No explicit observed data points are required
- The function's gradient (`∇F(p)`) and Hessian (`∇²F(p)`) are used for optimization
- This approach is more general and can represent a wider range of optimization problems
- Despite not having explicit data points, statistical calculations (standard errors, covariance) are still valuable, so a `degreeOfFreedom` parameter is needed to properly calculate these statistics

Both approaches use the same core Levenberg-Marquardt algorithm, but they require different interfaces and calculations.

### Proposed changes:

1. Add `IFixableObjectiveFunction` interface to allow fixing parameters during optimization
2. Rename existing `NonlinearObjectiveFunction` (internal class)  to `NonlinearObjectiveModel` for clarity
3. Add new `NonlinearObjectiveFunction` class for direct function minimization
4. Extend `LevenbergMarquardtMinimizer` to support `IObjectiveFunction`
5. Add `degreeOfFreedom` parameter for `IObjectiveFunction` to allow proper calculation of covariance and correlation matrices
6. Add `DirectMinimizationFunction` helper methods to `ObjectiveFunction` class

### Regarding `NonlinearFunction` methods:

The current `NonlinearFunction` methods create conceptual confusion because:

- They transform a model-based approach `f(x; p)` into a function-based approach `F(p)`
- This blurs the distinction between the two optimization approaches

The current implementation assumes a given `f(x; p)` is already `F(p)`, which is confusing.
The newly proposed `NonlinearLeastSquaresFunction` will properly convert a model function `f(x; p)` into an objective function `F(p) = Σ w_i * (y_i - f(x_i; p))²`.
I suggest marking the existing methods as deprecated to improve API clarity by:

- Maintaining clearer separation between model fitting and direct function minimization
- Directing users to either `NonlinearLeastSquaresFunction` or `NonlinearModel` based on their needs
- Providing `DirectMinimizationFunction` as a clearer alternative for direct function optimization

This extension would make the Levenberg-Marquardt algorithm more versatile, supporting both model fitting and direct function minimization while maintaining backward compatibility except for the deprecated methods.

I'd like to get your thoughts on the proposed deprecation of the current `NonlinearFunction` methods. Is this approach acceptable, or would you prefer a different solution to address the conceptual confusion?

I plan to submit a PR implementing these changes.

diluculo:
This PR implements:
- Support for direct residual functions in Levenberg-Marquardt algorithm
- Improved MINPACK-style control flow for damping parameter updates
- Comprehensive statistical metrics in `NonlinearMinimizationResult`
- Numerical differentiation through dedicated `NumericalJacobian` class

Closes #1114


diluculo:
Lastly, I've added a new `ParameterStatistics` class that computes all the key statistical measures for regression parameters, including standard errors, t-statistics, p-values, confidence intervals, and dependency analysis.

Ready for review and merge.

diluculo:
I've just added the `ILeastSquaresMinimizer` interface, which unifies our API for nonlinear least squares minimizers by providing two overloads for the `FindMinimum` method (one for Vector<double> and one for double[]). 

Please review and merge when ready. Thank you!

diluculo:
In the meantime, I found a critical bug in the Trust Region optimization implementation. The subproblem solvers were using unscaled gradient and Hessian values directly from the IObjectiveModel, while they should have been using the scaled values that NonlinearMinimizerBase.EvaluateJacobian produces.

The last commit fixes this issue by modifying the ITrustRegionSubproblem interface to accept the correctly scaled gradient and Hessian as parameters.

diluculo:
Summary of Key Changes

1. Enhanced Levenberg-Marquardt Algorithm
   - Support for direct residual functions
   - Improved MINPACK-style control flow for damping parameter updates

1. Comprehensive Statistical Analysis
   - New `ParameterStatistics` class for regression parameters analysis
   - Calculates standard errors, t-statistics, p-values, confidence intervals, and dependencies

1. API Improvements
   - Added `ILeastSquaresMinimizer` interface to unify nonlinear least squares minimizers

1. Bug Fixes
   - Fixed a bug in Trust Region optimization where subproblem solvers were using unscaled gradient and Hessian values
   - Modified `ITrustRegionSubproblem` interface to accept correctly scaled gradient and Hessian as parameters

febkor:
Hi @diluculo,
Thank you for the trying to improve the optimizers here.
There's a few independent changes in this PR which makes review difficult however.

Please could this bit be pulled out to a separate PR for now?
> Fixed a bug in Trust Region optimization where subproblem solvers were using unscaled gradient and Hessian values

diluculo:
Hi @febkor  

Thanks for your review. You're right, this PR contains several independent changes that should be separated. I've split it into 3 separate PRs:

1. Enhanced Levenberg-Marquardt Algorithm with residual support (including ParameterStatistics)
2. ILeastSquaresMinimizer interface : #1122 
3. Trust Region optimization bug fix : #1120
4. ~~ParameterStatistics for parameter inference~~

I initially planned to separate `ParameterStatistics` as well, but since it depends on changes to `IObjectiveModelEvaluation` in the LM enhancements, I've kept it in this PR to maintain functionality.

**Edit**:
In addition to splitting the PR, I've also added new features that I had been working on:
* The Basin Hopping global optimization algorithm in PR #1122
* The Exact Trust Region subproblem solver in PR #1120

Hi @cdrnet, 
This ended up being more work than initially anticipated, but I've completed everything I had prepared. I hope these PRs can be included in v6.0.

All PRs are now ready for review.

diluculo:
Regarding the failing test in AppVeyor: This is unrelated to the Levenberg-Marquardt enhancements - it's a floating-point precision issue in the statistics tests for .NET 8.0 (off by ~1.03E-09).

diluculo:
Hi @febkor 

Here is the mathematical background document I've prepared to explain the residual function support. It illustrates the relationship between the objective function (χ²) and the residual vector (R). I hope this helps in understanding the purpose and implementation of this PR.

image

src/Numerics/Optimization/NonlinearMinimizationResult.cs
```diff
@@ -31,19 +66,69 @@ public class NonlinearMinimizationResult
...
+        /// <param name="iterations">The number of iterations performed.</param>
+        /// <param name="reasonForExit">The reason why the algorithm terminated.</param>
         public NonlinearMinimizationResult(IObjectiveModel modelInfo, int iterations, ExitCondition reasonForExit)
         {
             ModelInfoAtMinimum = modelInfo;
             Iterations = iterations;
             ReasonForExit = reasonForExit;
 
             EvaluateCovariance(modelInfo);
+            EvaluateGoodnessOfFit(modelInfo);
```
febkor, line 125: Both EvaluateCovariance, EvaluateGoodnessOfFit seem like they should be optional.
The user might not need these, but they are relatively expensive to calculate.
On the master branch, EvaluateCovariance for example shows up in profiles as being quite significant (esp due to its allocations).

src/Numerics/Optimization/LevenbergMarquardtMinimizer.cs
```diff
@@ -140,91 +141,181 @@ public NonlinearMinimizationResult Minimum(IObjectiveModel objective, Vector<dou
...
+
+            var iterations = 0;
             while (iterations < maximumIterations && exitCondition == ExitCondition.None)
             {
                 iterations++;
 
                 while (true)
                 {
+                    // Store current Hessian diagonal for restoration if step is rejected
+                    var savedDiagonal = Hessian.Diagonal().Clone();
```
febkor, line 168: Is the separate `Clone()` required?
Matrix.Diagonal() produces a new vector with values copied in from the matrix.

src/Numerics/Optimization/LevenbergMarquardtMinimizer.cs
```diff
@@ -140,91 +141,181 @@ public NonlinearMinimizationResult Minimum(IObjectiveModel objective, Vector<dou
...
-            double mu = initialMu * diagonalOfHessian.Max(); // μ
-            double nu = 2; //  ν
-            int iterations = 0;
+            // Initialize trust region boundary delta and damping parameter mu
+            var delta = initialMu * P.L2Norm(); // Trust region boundary
+            if (delta == 0.0) delta = initialMu;
+            var mu = initialMu * Hessian.Diagonal().Max(); // Damping parameter μ
+            var nu = 2.0; // Multiplication factor ν for mu updates
+
+            // Counters for successful and failed iterations
```
febkor, line 153: The `ncsuc` counter seems to get incremented, but is not used for any logic?

src/Numerics/Optimization/ObjectiveFunctions/NonlinearObjectiveModel.cs
```diff
@@ -0,0 +1,688 @@
...
+            evaluationCount = 0;
+
+            // Process each observation point separately
+            for (var i = 0; i < NumberOfObservations; i++)
+            {
+                var obsIndex = i; // Capture observation index for the lambda
+
+                // Create adapter function that returns the model value for current observation
+                // when given parameters array
+                double funcAdapter(double[] p)
```
febkor, line 580: It seems it'd be more natural to extend the NumericalJacobian to support modelFunctions of the shape needed here?
The usage here seems more complex than needed and some inefficiencies:
- _modelFunction returns a whole vector, but only a single value is used
- this funcAdapter lambda creates an allocation per observation (minor in the context of the rest of the code, but still avoidable)

src/Numerics/Optimization/ObjectiveFunctions/NonlinearObjectiveModel.cs
```diff
@@ -0,0 +1,688 @@
...
+            if (weights != null && weights.Count(x => x == 0) == weights.Count)
+            {
+                throw new ArgumentException("All the weightings can't be zero.");
+            }
+            if (weights != null && weights.Count(x => x < 0) > 0)
+            {
+                weights = weights.PointwiseAbs();
+            }
+
+            Weights = (weights == null)
```
febkor, line 364: Does the `Weights` matrix get used anywhere?
Apart from some trivial null checks, the usages I noticed all only use the diagonal, i.e. the `weights` vector. Are there any I might have missed?
It can be quite expensive in memory terms to create it.

febkor (commented):
Just a brief review, more from a performance angle.
Hope you find them useful.

PR #1120, opened 2025-04-02
Fix and Extend Trust Region Subproblem Implementations

diluculo:
This PR fixes critical issues in the Trust Region optimization implementation.

### Bugs Fixed
1. **Scaling Issue**: Subproblem solvers were using unscaled gradient and Hessian values directly from the `IObjectiveModel`, while they should have been using the scaled values that `NonlinearMinimizerBase.EvaluateJacobian` produces.

2. **NewtonCG Algorithm Improvement**: Enhanced the handling of negative curvature directions in the NewtonCG subproblem solver.

### DIFF
--- a/src/Numerics.Tests/OptimizationTests/NonLinearCurveFittingTests.cs
+++ b/src/Numerics.Tests/OptimizationTests/NonLinearCurveFittingTests.cs
@@ -202,6 +202,20 @@ namespace MathNet.Numerics.Tests.OptimizationTests
         }
 
         [Test]
+        public void Rat43_TREXACT_Dif()
+        {
+            var obj = ObjectiveFunction.NonlinearModel(Rat43Model, Rat43X, Rat43Y, accuracyOrder: 6);
+            var solver = new TrustRegionExactMinimizer();
+            var result = solver.FindMinimum(obj, Rat43Start2);
+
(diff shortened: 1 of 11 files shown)


diluculo:
Regarding the failing test in AppVeyor: This is unrelated to the Trust Region fix - it's a floating-point precision issue in the statistics tests for .NET 8.0 (off by ~1.03E-09).

diluculo:
I'd like to extend this PR by adding an implementation of the Exact Trust Region Subproblem solver. I initially developed this solver some time ago for my own work, and I've now updated it to be compatible with the bug fixes introduced in this PR.

Comparison of the Three Trust Region Subproblem Solvers:
* **Dog-Leg**: Provides a good trade-off between computational efficiency and convergence speed, especially when the iterate is already close to the solution. It performs best when the Hessian is positive definite, but it cannot handle indefinite matrices or negative curvature.
* **Newton-CG**: More robust than Dog-Leg. This method handles indefinite Hessians effectively and detects negative curvature directions. It is well-suited to large-scale problems due to its iterative structure and better scalability.
* **Exact**: Solves the trust region subproblem via eigenvalue decomposition, providing the most accurate solution. Although it comes with a higher computational cost, it is the best choice when high precision is required or when the Hessian has negative eigenvalues. It is the most theoretically complete approach among the three.

PR #1139, opened 2025-08-26
feat: Improve intel MKL shared library build process using docker

sudiptob2:
MathNet.Numerics supports **Intel MKL** for CPU acceleration. On **Linux**, MKL support relies on the native provider **`MathNet.Numerics.MKL.Linux`**. Currently, this provider is not bundled with **MathNet.Numerics** due to size constraints. Instead, it is available as a separate **NuGet** package. However:

- The **NuGet** package is **outdated**. [Please take a look at this discussion.](https://github.com/mathnet/mathnet-numerics/discussions/1133)
- It is **incompatible** with the latest builds from the `master` branch.

To work with **MKL**, users must build the provider manually, which involves compiling the **Intel shared libraries** using the [**`mkl_build.sh`**](https://github.com/mathnet/mathnet-numerics/blob/master/src/NativeProviders/Linux/mkl_build.sh) script. This script is **outdated** and depends on pre-installed **host machine dependencies**.

To streamline the process, in this PR following is proposed:

1. **Updating** the build script for **Ubuntu 24.04**.
2. Adding a **Dockerfile** to build the shared libraries inside a container, **eliminating host dependency requirements**.
3. Improving **documentation** on how to use the script and Dockerfile.


sudiptob2:
Hi @cdrnet, any thoughts on this?

PR #1140, opened 2025-10-14
upd

is-m-34-ail:


jkalias (changes requested):
What exactly is this change supposed to improve in the code base?

PR #1141, opened 2025-11-17
Add trimming attributes

Cheesebaron:
To support scenarios where application have trimming enabled, I've annotated the code with trimming attributes.

There is still a warning produced by `var type = Type.GetType(typeName);` which complains about it cannot guarantee the `typeName` is available. Which is fine I guess.


dartasen:
@cdrnet Sorry for the ping, love the library and using it for mobile apps that requires full trimming, any chance this Merge Request could be reviewed/merged at some point ? Thanks 😄

